PR #57, opened 2021-09-19
Notifications

verdefred:


app/models/notifications/coin_bought.rb
```diff
@@ -0,0 +1,4 @@
+module Notifications
+  class CoinBought < Notification
```
francisco-leal, line 2: Let's change this name to "TokenAcquired" we're calling it Talent Tokens and no longer Career Coins

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,23 @@
+class NotificationsController < ApplicationController
```
francisco-leal, line 1: Since this endpoint is just used internally by a React component let's change it to the API namespace

app/packs/stylesheets/bootstrap/_overrides.scss
```diff
@@ -21,5 +21,7 @@ $pagination-active-color: $primary !default;
 $pagination-active-bg: none !default;
 $pagination-disabled-bg: none !default;
 
+$border-color: #dee2e6 !default;
```
francisco-leal, line 24: this might change a lot of things in the app since borders are used very widely. Can we instead use the already existing $border-color that bootstrap provides so we don't mess with that much styling? Also if you want bootstrap already provides a greyscale that we can use.

app/services/CreateUser.rb
```diff
@@ -57,6 +57,7 @@ def create_user(email, username, password)
 
   def create_talent(user)
     user.create_talent!
+    CreateNotificationNewTalentListedJob.perform_later
```
francisco-leal, line 60: I think we should make this notification once the profile is set as public and now when we create the user. I think it's more interesting that we notify when the profile is ready. What do you think?

app/services/UpdateTalent.rb
```diff
@@ -54,4 +56,11 @@ def update_tags(tags)
       tag.save!
     end
   end
+
+  def create_notification_talent_changed(user)
+    service = CreateNotification.new
+    user.followers.find_each do |u|
+      service.call(body: "#{user.display_name} changed", user_id: u.id, type: 'Notifications::TalentChanged')
```
francisco-leal, line 63: Maybe the better message is "#{user.display_name} profile has been updated"

app/services/CreateUser.rb
verdefred: Totally agree, I'll make the change. Good thinking, thanks!

app/services/UpdateTalent.rb
verdefred: To keep this generic, maybe I'll remove the name of the user. This way when we create a notification when a talent has changed, we show a generic message. If another talent makes a change after the first one we don't create a new notification (if the first one is not read). What do you think?

verdefred: After talking IRL we decided to go with the display name and redirect the user directly to the talent changed page

francisco-leal (approved):

PR #135, opened 2021-11-20
Tooltip

verdefred:


andreasvilela (approved):

app/packs/src/components/design_system/tooltip/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+      <Popover
+        id={popOverAccessibilityId}
+        className={cx(className, 'talent-tooltip', mode)}
+      >
+        {popOverContent ?
+          popOverContent :
+          <>
+            {title && <P3 text={title} bold className="tooltip-title"/>}
+            {body && <P3 text={body} bold className="tooltip-body"/>}
+          </>
```
francisco-leal, line 34: Nice! Just to be 100% clear we either pass popOverContent or we pass title + body right?

francisco-leal (approved):

app/packs/src/components/design_system/tooltip/index.jsx
verdefred: You can also only pass the body, with no title, as per the designs
image

francisco-leal: nice!

PR #140, opened 2021-11-24
#textinput field

andreasvilela:


.vs/VSWorkspaceState.json
```diff
@@ -1,16 +1,9 @@
 {
```
francisco-leal, line 1: We should add rules to ignore these files to the `.gitignore`a rule like `.vs/` (similar to the already existing one for `.vscode/` should work

app/packs/stylesheets/custom/_fields.scss
```diff
@@ -0,0 +1,87 @@
+@import "../colors";
```
francisco-leal, line 1: There is one pattern in this file that we should change. We shouldn't hardcore hex colors, it will make it very hard if we want to change our colour palette in the future, we should try to always use css variables. If they don't exist yet let's create them or if they are very close to an already existing one we should check why they are different

.vs/VSWorkspaceState.json
andreasvilela: Added. Thank you.

app/packs/stylesheets/custom/_fields.scss
andreasvilela: Ok, will change that, reviewing if already exists, otherwise create new ones.

PR #144, opened 2021-11-25
Av/design system

andreasvilela:
Fields textinput & Alerts


app/packs/src/components/design_system/alert/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+                    {type === 'info' && <Info />}
+                    {type === 'critical' && <Critical />}
+                </div>
+                <div className="alertRightArea">
+                    <strong>{title}</strong> 
+                    <p>{message}</p>
+
+                    {link ? <a href={`${link}`} target="_blank">Go to link page</a> : null}
+
+                    <a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>
```
francisco-leal, line 31: `class` should be `className` here

app/packs/src/components/design_system/alert/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+                    {type === 'success' && <Success />}
+                    {type === 'warning' && <Warning />}
+                    {type === 'info' && <Info />}
+                    {type === 'critical' && <Critical />}
+                </div>
+                <div className="alertRightArea">
+                    <strong>{title}</strong> 
+                    <p>{message}</p>
+
+                    {link ? <a href={`${link}`} target="_blank">Go to link page</a> : null}
```
francisco-leal, line 29: We probably don't want the "Go to link page" to be hardcoded right?

app/packs/src/components/design_system/alert/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+    mode,
+    title,
+    message,
+    link,
+}) => {
+
+    return (
+        <div className="col-md-4">
+            <div className={`alert ${mode} alert-${type} alert-dismissible d-flex`}>
+                <div>
```
francisco-leal, line 19: If this `<div>` doesn't have any class or any behaviour then I don't think we need it

app/packs/src/components/design_system/fields/textinput/index.jsx
```diff
@@ -0,0 +1,23 @@
...
+    mode,
+    disabled,
+}) => {
+
+    return (
+        <div>
+            <div>
+                <h6 className={`title-field ${mode}`}>{title}</h6>
+            </div>
+            <input type="text" className={`form-control ${mode}`} placeholder={placeholder} disabled={disabled} />
```
francisco-leal, line 16: We need a way to pass the value of the input text to outside of this component, otherwise we won't be able to.

app/packs/src/components/design_system/fields/textinput/index.jsx
```diff
@@ -0,0 +1,23 @@
...
+    shortCaption,
+    placeholder,
+    mode,
+    disabled,
+}) => {
+
+    return (
+        <div>
+            <div>
+                <h6 className={`title-field ${mode}`}>{title}</h6>
```
francisco-leal, line 14: Will all `TextInput` components have a Title and a shortCaption? If not we should only add them to the html if they are passed to the component

app/packs/src/components/design_system/fields/textinput/index.jsx
```diff
@@ -0,0 +1,23 @@
...
+    title,
+    shortCaption,
+    placeholder,
+    mode,
+    disabled,
+}) => {
+
+    return (
+        <div>
+            <div>
```
francisco-leal, line 13: If these divs don't have any classes or behaviours then we don't need them. If we need to group the html because the component needs to return a single element we can use `<>` `</>`, they won't add anything to HTML but for they will signal react to treat them as a single entity

PR #145, opened 2021-11-28
Checkbox

verdefred:


app/packs/src/components/design_system/checkbox/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+const Checkbox = ({ checked, onChange, disabled, label, mode }) => (
+  <label className="container-checkbox">
+    <input
+      type="checkbox"
+      checked={checked}
+      onChange={onChange}
+      disabled={disabled}
+    />
+    <span className="checkmark"></span>
+    {label && <P3 text={label} mode={mode} />}
```
francisco-leal, line 14: only the text changes based on the mode?

app/packs/src/components/design_system/slider/index.jsx
```diff
@@ -1,22 +1,22 @@
 import React from "react";
-import PropTypes from "prop-types";
+import { func, bool } from "prop-types";
 
-const Slider = (props) => {
+const Slider = ({ onClick, checked }) => {
   return (
     <label className="form-switch">
-      <input type="checkbox" checked={props.checked} onChange={() => props.onClick()} />
+      <input type="checkbox" checked={checked} onChange={() => onClick()} />
```
francisco-leal, line 7: `onChange={onClick}` would be even better because then we could have access to the event if whoever is using the input needs it

francisco-leal (approved):
Looks good! left some minor comments

app/packs/src/components/design_system/checkbox/index.jsx
verdefred: Yup, exactly! But I did ask for more clarity on the designs for when the checkbox is disabled and our great designer quickly added it 💪 
I'll be applying those changes this week

app/packs/src/components/design_system/slider/index.jsx
verdefred: Great catch, I'll make the change. Thanks!

PR #162, opened 2021-12-13
Table

verdefred:


stories/components/table/Table.stories.mdx
```diff
@@ -0,0 +1,155 @@
...
+
+Custom Tables.
+
+## Examples
+
+export const Template = (args) => (
+  <Table {...args}>
+    <Table.Head>
+      <Table.Th>
+        <Caption bold text="Supporter" />
```
verdefred, line 30: Should this <Caption> go inside the Th? We can change the Caption (and P1, P2, ...) to accept a children instead of only a text

andreasvilela: I do not have a specific comment on this. What do you think @ikrakel ?

francisco-leal: We'll need the table header to be clickable to enable filtering and sorting through the different columns, so we'll always need the header cells to receive custom children

verdefred: Good point, makes sense. I'll keep it as is 💪 
What to you think about the rest of the component? All good?

PR #167, opened 2021-12-14
Av/design system

andreasvilela:


app/packs/src/components/design_system/fields/dropdown/index.jsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+const TextInput = ({ title, shortCaption, mode, disabled, options }) => {
```
francisco-leal, line 3: we need callbacks when items are selected so that we can know which value is selected. Also let's use the Typography componentes instead of plain <h6> elements, same for the short caption.

app/packs/src/components/design_system/fields/dropdown/index.jsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <>
+      {title ? <h6 className={`title-field ${mode}`}>{title}</h6> : null}
+
+      <select className={`form-control ${mode}`} disabled={disabled}>
+        {options.map((option) => (
+          <option value={option.value}>{option.label}</option>
+        ))}
```
francisco-leal, line 11: Do we have empty states on these dropdowns? what happens when no options exist?

app/packs/src/components/design_system/modal/contact.jsx
```diff
@@ -21,7 +22,78 @@ const Modal = ({ mode }) => {
...
+              text="Supporting"
+            ></Caption>
+            <div className="mt-1">
+              <Avatar
+                name="John Travolta"
+                imgUrl="https://i.pravatar.cc/150?img=1"
+                size="big"
+                mode={`${mode}`}
+              />
+            </div>
```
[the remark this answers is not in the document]
andreasvilela: Yes, we can do that way. Changed Avatar to support className prop. ;)

app/packs/src/components/design_system/fields/dropdown/index.jsx
andreasvilela: What's your suggestion on this ?

francisco-leal: I'm not sure we should check with Pedro Pereira

andreasvilela: Will ask him today during the call.

app/packs/src/components/design_system/fields/dropdown/index.jsx
andreasvilela: Done.

francisco-leal (approved):

PR #182, opened 2021-12-29
Admin create talent

verdefred:


app/services/talent/create_talent.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class Talent::CreateTalent
+  def initialize
```
RubenSousaDinis, line 4: redundant initializer, can be removed

RubenSousaDinis (approved):
In terms of code looks good to me. Just left a minor comment

PR #208, opened 2022-01-18
Redirect to user profile on user avatar click

RubenSousaDinis:
Clicking on messages user avatar is now redirecting to the user profile page.

From now on passing the link prop to TalentProfilePicture will create a link to the desired path.

Tested both desktop and mobile.

### DIFF
--- a/app/packs/src/components/talent/TalentProfilePicture.jsx
+++ b/app/packs/src/components/talent/TalentProfilePicture.jsx
@@ -14,6 +14,7 @@ const TalentProfilePicture = ({
   straight,
   blur,
   border,
+  link,
 }) => {
   const [currentTheme, setCurrentTheme] = useState(document.body.className);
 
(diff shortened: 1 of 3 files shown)


app/packs/src/components/talent/TalentProfilePicture.jsx
```diff
@@ -10,22 +10,25 @@ const TalentProfilePicture = ({
...
-    <img
-      className={`${roundPhoto}${grey}${blurPhoto}${borderPhoto} ${
-        className || ""
-      } image-fit`}
-      src={imgSrc}
-      width={height}
-      height={height}
-      alt="Profile Picture"
-    />
+    <a href={link}>
```
verdefred, line 21: can we put the `a` only when a link is passed? in some cases we might want just the photo. with the `a` as a wrapper we'll always have link properties, like the cursor: pointer, for example. that may be confusing since without a link the photo is not clickable. Other than this we should also add `link: string` to the propTypes

verdefred (commented):

app/packs/src/components/talent/TalentProfilePicture.jsx
RubenSousaDinis: good point about the pointer. Will address that

RubenSousaDinis: Fixed!

verdefred (approved):
Looks great 🎉

PR #211, opened 2022-01-18
Ignore .env from github. Add env sample file

RubenSousaDinis:
Add `.env` to gitignore since we need to update it to run the app locally.

Added `.env.sample` with an example of necessary env vars to run the app


RubenSousaDinis:
Just noticed we're using `.env.development`

PR #215, opened 2022-01-19
Fix overview challenges and pitch and token details

francisco-leal:


app/packs/src/components/talent/Show/Overview.jsx
```diff
@@ -14,6 +14,7 @@ import Divider from "../../design_system/other/divider";
 import { P1, P2, H5, Caption } from "src/components/design_system/typography";
 
 const Overview = ({ sharedState, mode }) => {
+  console.log(sharedState);
```
verdefred, line 17: extra console.log

app/packs/src/components/talent/Show/TokenDetails.jsx
```diff
@@ -18,7 +18,7 @@ const TokenDetails = ({
   displayName,
   username,
   mobile,
-  className,
+  className = "",
```
verdefred, line 21: do we need a default?

francisco-leal: it was adding an `undefined` as class name

verdefred (approved):

PR #223, opened 2022-01-21
UI edit profile improvements

verdefred:
UI edit profile settings tab improvements


app/controllers/users_controller.rb
```diff
@@ -16,7 +16,7 @@ def index
   end
 
   def create
-    if verify_captcha
+    if verify_captcha && User.valid_username?(user_params[:username])
```
francisco-leal, line 19: don't we need a different error message for this?

verdefred: nice catch, will add 👍

francisco-leal (approved):

PR #224, opened 2022-01-21
Send message to all supporters

RubenSousaDinis:
This PR adds the message to all supporters endpoint.

It adds a job to send messages to all supporters. We've used `sidekiq-status` gem to track the progress of the messages. (https://github.com/kenaniah/sidekiq-status).

In the Frontend we're pulling every 2 seconds to check if more messages were sent.

In the video below I've added a 3 sleep in the job to test:

https://www.loom.com/share/75588bdbc4224675a028aaed004e7ede


RubenSousaDinis:
@francisco-leal this might require some more tweaks in the UI. A success message with the number of messages sent or the list of supporters in the modal before sending might be good to give more information to the user

app/packs/src/components/chat/MessageUserList.jsx
```diff
@@ -1,8 +1,11 @@
 import React, { useState, useEffect } from "react";
+import { faSearch } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
```
verdefred, line 3: I think this can be removed

app/packs/src/components/chat/NewMessageModal.jsx
```diff
@@ -65,6 +66,12 @@ const NewMessageModal = ({ show, setShow, onUserChosen, mobile }) => {
             className="position-absolute chat-search-icon"
           />
         </div>
+        <div className="w-100 d-flex flex-column px-4 py-2">
+          <a onClick={() => setShowMessageToAllSupporters()}>
+            <Send color="currentColor" className="mr-2" />
+            Send message to all supporters
```
verdefred, line 72: this should be
```jsx
<P2 className="text-black" text="Send message to all supporters" bold />
```

app/packs/src/components/chat/NewMessageModal.jsx
```diff
@@ -65,6 +66,12 @@ const NewMessageModal = ({ show, setShow, onUserChosen, mobile }) => {
             className="position-absolute chat-search-icon"
           />
         </div>
+        <div className="w-100 d-flex flex-column px-4 py-2">
+          <a onClick={() => setShowMessageToAllSupporters()}>
```
verdefred, line 70: for the `currentColor` of the `<Send />` to work, we should add a className="text-primary" here. Also, since this is a onClick, it probably should be a button. we can change to something like:
```jsx
<button className="button-link text-primary" onClick={() => setShowMessageToAllSupporters()}>
```

app/packs/src/components/chat/NewMessageToAllSupportersModal.jsx
```diff
@@ -0,0 +1,67 @@
+import React, { useState, useEffect } from "react";
+import Modal from "react-bootstrap/Modal";
+import TextArea from "src/components/design_system/fields/textarea";
+import debounce from "lodash/debounce";
+import { post } from "src/utils/requests";
+
+const NewMessageToAllSupportersModal = ({ show, setShow, mode, mobile }) => {
+  if (!show) {
+    return null;
+  }
```
verdefred, line 10: Do we need this?

app/packs/src/components/chat/NewMessageToAllSupportersModal.jsx
```diff
@@ -0,0 +1,67 @@
...
+            rows="5"
+          />
+
+        <button
+          onClick={debouncedNewMessage}
+          disabled={sendingMessage}
+          className="btn btn-primary talent-button primary-default-button bold w-40 mt-2 float-right"
+        >
+          Send message
+        </button>
```
verdefred, line 61: here it's easier if we use our custom Button. it would be something like:
```jsx
<Button
  type="primary-default"
  onClick={debouncedNewMessage}
  disabled={sendingMessage}
  className="mt-2 float-right"
/>
```

app/services/messages/send_to_all_supporters.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      investors.map do |investor|
+        receiver = investor.user
+
+        message = send_message_to(receiver)
+        create_notification_for(receiver)
+        broadcast(message)
+
+        message
+      end
```
verdefred, line 23: should we have this wrapped in a ActiveRecord::Base.transaction? Not sure if we want to not send the message to anyone if some individual message fails. @francisco-leal can you give your opinion please?

spec/requests/messages_spec.rb
```diff
@@ -0,0 +1,73 @@
+require "rails_helper"
```
verdefred, line 1: Should the name of this file be messages_controller_spec.rb?

verdefred (commented):
Great work overall, just left some comments 👏

spec/requests/messages_spec.rb
RubenSousaDinis: This is not a controller spec. It's a request spec which means that we're not testing a particular class here but a set of requests within the same path. In this case the path is `/messages`.

app/services/messages/send_to_all_supporters.rb
RubenSousaDinis: I believe we may want to move this into a Sidekiq job. We can have talents with a huge set of supporters which will lead to a slow response.

app/packs/src/components/chat/NewMessageToAllSupportersModal.jsx
RubenSousaDinis: It's just a guard clause to ensure the modal will never render if show is not true. It should never happen tho.

verdefred: I think it can be removed because it'll never happen 👍

app/packs/src/components/chat/MessageUserList.jsx
RubenSousaDinis: yup, left-over from conflicts

app/packs/src/components/chat/NewMessageToAllSupportersModal.jsx
RubenSousaDinis: Done!

PR #239, opened 2022-01-27
Add pre commit hook to run linter before each commit

RubenSousaDinis:
This PR adds a pre-commit hook that runs standardrb for staged files.

In order to setup the hook you will need to run the following command in the repo root folder:

`cp .githooks/pre-commit .git/hooks` 

Example output:

Screenshot 2022-01-27 at 21 16 46

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,37 @@
+#!/bin/sh
+#
+# Check for ruby style errors
+
+red='\033[0;31m'
+green='\033[0;32m'
+yellow='\033[0;33m'
+NC='\033[0m'
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+  against=HEAD
+else
+  # Initial commit: diff against an empty tree object
+  # Change it to match your initial commit sha
+  against=744c04f2b1ad7aaa24ea0df7831680f7ddddbe58
+fi
+
+# Check if standardrb is installed for the current project
+bin/bundle exec standardrb -v >/dev/null 2>&1 || { echo >&2 "${red}[Ruby Style][Fatal]: Add standardrb to your Gemfile"; exit 1; }
+
+echo "${green}[Ruby Style][Info]: Checking Ruby Style${NC}"
+
+echo "${green}[Ruby Style][Info]: ${FILES}${NC}"
+
+# Run standardrb on the staged files
+output=$(make lint)
+
+if [ $? -ne 0 ]; then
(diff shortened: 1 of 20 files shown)


Makefile
```diff
@@ -0,0 +1,2 @@
+lint: 
+	git diff --name-only --cached --diff-filter=d | grep '**/*\.rb\|**/*\.rake\|Gemfile' | grep -v 'Gemfile.lock' | xargs bundle exec standardrb --extra-details --fail-level A
```
RubenSousaDinis, line 2: This is the command that runs the linter for staged files only.

.githooks/pre-commit
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
RubenSousaDinis, line 1: This is the pre-commit script.

pdfrod:
Nice! I just have a couple of comments:

1. There's still one pending warning when running the linter. Shouldn't we try to fix it, or at least add it to the TODO file? Screenshot_2022-01-27_22-30-24
2. What do you think of adding the linter instructions (i.e., `cp .githooks/pre-commit .git/hooks` and running `make lint`) to the README.md file?

francisco-leal:
@pdfrod I think there is no way around it as Redis uses that global variable itself so I'm not sure we have an alternative. I'm sure there is a way to ignore however so that we don't get a warning

francisco-leal:
@RubenSousaDinis I agree with @pdfrod, if we can add instructions to the README we'll be ready to ship this!

RubenSousaDinis:
> @pdfrod I think there is no way around it as Redis uses that global variable itself so I'm not sure we have an alternative. I'm sure there is a way to ignore however so that we don't get a warning

I believe that the initializer can be removed. We're using heroku which means that we just need to point our app to the correct redis instance URL. Locally we can do the same and point to our local redis server.

RubenSousaDinis:
> Nice! I just have a couple of comments:
> 
> 1. There's still one pending warning when running the linter. Shouldn't we try to fix it, or at least add it to the TODO file? Screenshot_2022-01-27_22-30-24
> 2. What do you think of adding the linter instructions (i.e., `cp .githooks/pre-commit .git/hooks` and running `make lint`) to the README.md file?

Readme notes added. @francisco-leal I typed a bit more about the git approach, feel free to change it.

verdefred (approved):
Pushed to our dev env and working great 🚀

PR #289, opened 2022-02-10
Create discovery page

verdefred:


app/controllers/discovery_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    @launching_soon_talents = base_talent
+      .where(token: {contract_id: nil})
+      .order(created_at: :desc)
+      .limit(3)
+
+    @discovery_rows = []
+
+    DiscoveryRow.find_each do |row|
+      talents_by_tag = base_talent.joins(:tags).where(tags: {id: row.tags.pluck(:id)}).distinct.order(:id)
+      @discovery_rows << {title: row.title, talents: talents_by_tag.sample(4, random: Random.new(Date.today.jd))}
```
francisco-leal, line 19: `sample` is an Array method which means that we are loading all Talent into memory here which might not be scalable. Is it possible to do this using just SQL?

app/models/talent_tag.rb
```diff
@@ -1,4 +1,15 @@
 class TalentTag < ApplicationRecord
   belongs_to :talent
   belongs_to :tag
+
+  validate :one_talent_tag_combination
+
+  private
+
+  def one_talent_tag_combination
```
francisco-leal, line 9: Isn't it better to add a uniqueness constraint to the database with these two fields? Rails validations don't actual guarantee uniqueness

app/packs/src/components/design_system/cards/NewTalentCard.jsx
```diff
@@ -98,15 +122,18 @@ const NewTalentCard = ({
...
+  isFollowing: bool.isRequired,
   updateFollow: func.isRequired,
   talentLink: string.isRequired,
 };
 
-export default NewTalentCard;
+export default (props) => (
+  <ApolloProvider client={client(props.railsContext.contractsEnv)}>
+    <NewTalentCard {...props} />
+  </ApolloProvider>
```
francisco-leal, line 138: This is an issue, this means that we'll have x times the number of rendered cards using the apollo provider and each of them asking data from the graph. It's better if we have the provider on the page component and pass the values required to each card.

app/controllers/discovery_controller.rb
```diff
@@ -0,0 +1,36 @@
+class DiscoveryController < ApplicationController
+  def index
```
francisco-leal, line 2: considering we don't always show the latest added, launching soon should we move those to a separate controller that react calls once it renders? instead of it being on this method?

app/packs/src/components/discovery/index.jsx
```diff
@@ -0,0 +1,238 @@
...
+  discoveryRows,
+  marketingArticles,
+}) => {
+  const { mobile } = useWindowDimensionsHook();
+  const [localMostTrendyTalents, setLocalMostTrendyTalents] = useState([]);
+  const [localLatestAddedTalents, setLocalLatestAddedTalents] =
+    useState(latestAddedTalents);
+  const [localDiscoveryRows, setLocalDiscoveryRows] = useState(discoveryRows);
+
+  const addTokenDetails = (talents, talentsFromChain) => {
```
francisco-leal, line 33: should we memoize this?

lib/tasks/tmp/add_basic_role_to_users.rake
```diff
@@ -0,0 +1,7 @@
+namespace :users do
+  task add_basic_role: :environment do
+    User.where(role: nil).find_each do |user|
+      user.update(role: "basic")
```
francisco-leal, line 4: We'll also need to update the core team roles or are we considering it would already be done before we merge this?

verdefred: I was thinking of changing before this, what do you think?

verdefred: Resolving since I already did this

francisco-leal (approved):

PR #306, opened 2022-02-21
Use noticed gem

pdfrod:
### What changed?

Notifications are now implemented using [Noticed](https://github.com/excid3/noticed) gem and the "Talent listed" notification was removed. From the user perspective, the only visible change is that the "Talent listed" notification doesn't exist anymore; everything else should be exactly the same.

### Why it changed?

Noticed includes functionality to easily send notifications by email (which we'll need) and also makes it easy to add new notification types.

### How it changed?

- Fields were added and removed from the `notifications` database table to match Noticed expect fields. The old fields were migrated to the new format.
- The existing Notification models were replaced by other classes deriving from Noticed class hierarchy.
- Notification text is now inferred from the notification type at display time, instead of being stored in the notification.

## Notion link

[Notification System](https://www.notion.so/talentprotocol/Notification-System-4a4d3446ca8444db80dcedbe46291f3e)

## How to test?

- Run the database migration
- Confirm that the existing notifications are working correctly
- Confirm that new notifications are created correctly

## Notes

The database migration makes backward incompatible changes to the schema, so the application should be put on maintenance mode during the deployment.


francisco-leal (commented):
Initial review looks good! I'll checkout the branch to take a look at the migration with more details

francisco-leal (dismissed):
This looks good. We have some tests running in the dev environment so don't merge yet but the migration tot he noticed gem is fine 👍

app/controllers/api/v1/notifications_controller.rb
```diff
@@ -1,23 +1,15 @@
...
-  before_action :set_notification, only: [:update]
-
-  def update
-    result = @notification.update(notification_params)
-
-    if result
-      render json: {success: "Notification successfully updated."}, status: :ok
+  def mark_as_read
+    id = params[:notification_id]
+    notification = current_user&.notifications&.find_by_id(id)
```
verdefred, line 4: `find_by_id` is deprecated, please use `find_by(id: id)`
https://edgeguides.rubyonrails.org/4_0_release_notes.html#active-record-deprecations 
Also, since this is an api request, I don't think there's ever a situation where current_user doesn't exist, so we can remove the first `&`

app/notifications/base_notification.rb
```diff
@@ -0,0 +1,34 @@
...
+  def body
+    t(".body", name: source_name)
+  end
+
+  def source
+    if @source_loaded
+      @source
+    else
+      @source_loaded = true
+      @source = User.find_by_id(params["source_id"])
```
verdefred, line 13: Same here regarding find_by instead of find_by_id

app/services/create_notification.rb
```diff
@@ -1,14 +1,26 @@
-class CreateNotification
-  def initialize
-  end
-
-  def call(title:, body:, user_id:, type:, source_id: nil)
-    notification = Notification.where(user_id: user_id, type: type, read: false, source_id: source_id).last
+module CreateNotification
+  def self.call(recipient:, type:, source_id: nil)
```
verdefred, line 2: Is there a reason for defining a class method instead of an instance method, here? If possible I think we should be consistent in the way we do this, in this case creating services. In all other services we define service methods as instance methods

verdefred (commented):
Looks good overall, just left a couple of comments

app/controllers/api/v1/notifications_controller.rb
RubenSousaDinis: Why not using find? It will raise 404 to the FE which I believe is the right thing to do here

config/routes.rb
```diff
@@ -39,7 +39,9 @@
         resources :users, only: [:index, :show, :update, :destroy]
         resources :follows, only: [:index, :create]
         delete "follows", to: "follows#destroy"
-        resources :notifications, only: [:update]
+        resources :notifications, only: [] do
+          post :mark_as_read
```
RubenSousaDinis, line 43: Looking at the `mark_as_read` action it's not creating any resource in the server, so I believe we should keep it as a put/patch, request. What do you think?

db/migrate/20220205193302_change_notifications.rb
```diff
@@ -0,0 +1,54 @@
...
+  def change
+    add_reference :notifications, :recipient, polymorphic: true
+    add_column :notifications, :params, :jsonb
+    add_column :notifications, :read_at, :datetime
+    add_index :notifications, :read_at
+
+    reversible do |dir|
+      dir.up do
+        execute <<~SQL
+          DELETE FROM notifications WHERE type = 'Notifications::TalentListed';
```
RubenSousaDinis, line 11: Not a big fan of data migrations inside schema migrations. There're some ways we can manage data migrations:

- Writing a rake task
- Using something like this: https://github.com/OffgridElectric/rails-data-migrations

Feel free to proceed with it as is if others are happy with it. If we agree to proceed let's try to check the amount of notifications we have in prod and deploy it off peak hours to prevent database locking at deploy.

app/controllers/api/v1/notifications_controller.rb
pdfrod: Good points! I'll replace `find_by_id` by `find` and remove the first `&`.

app/notifications/base_notification.rb
```diff
@@ -0,0 +1,34 @@
+class BaseNotification < Noticed::Base
+  deliver_by :database
+
+  def body
+    t(".body", name: source_name)
+  end
+
+  def source
+    if @source_loaded
```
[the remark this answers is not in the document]
pdfrod: Can't use exactly the same as in `source_name`, because if the User deleted their account then `@source` will be `nil` and we'd have no way to know if we already tried to load the user or not.

app/notifications/base_notification.rb
```diff
@@ -0,0 +1,34 @@
...
+      else
+        source.username
+      end
+  end
+
+  def title
+    t(".title")
+  end
+
+  def url
```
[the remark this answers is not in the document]
pdfrod: Yes, good point.

app/services/create_notification.rb
pdfrod: Well this module doesn't have internal state, so it felt like a waste of memory and CPU cycles to allocate an object just to do a function call and discard it right after. But sure I can turn into a class for consistency.

app/services/create_notification.rb
```diff
@@ -1,14 +1,26 @@
...
+      notification = type.with(params)
+      notification.deliver_later(recipient)
+    end
+  end
+
+  def self.get_existing_unread(recipient:, type:, source_id:)
+    notifications = Notification.where(type: type.name, recipient: recipient,
+                                       read_at: nil)
+    if source_id
+      notifications = notifications.where("(params->>'source_id')::bigint = ?",
```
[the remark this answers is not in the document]
pdfrod: The normal way to add additional data using Noticed gem seems to be using the `params` object. Trying to make the `source_id` column play nicely with Noticed might require some hacks, some I'm not sure it would make things any better. But yeah at least a scope sounds great, I'll do that.

db/migrate/20220205193302_change_notifications.rb
pdfrod: Definitely something to explore in future migrations, but using a data migration here is not that easy because there are schema changes that can only be applied after the data migration. So I'd need to create two schema migrations and one database migration, plus I'd have to coordinate the data migration to make sure it runs after the first schema migration and before the second schema migration.

The schema changes are not backwards compatible, so a bit of downtime will be required anyway. I could have tried to avoid downtime, but it would complicate things considerably. After discussing with @francisco-leal we agreed that at this stage avoiding downtime is not that critical, and that our time is probably better spent elsewhere.

config/routes.rb
pdfrod: Since it's an idempotent request, I guess a `PUT` method would be fitting, yes.

pdfrod: Oh wait, apparently it's not idempotent at the moment - the `read_at` gets updated with the current date every time this endpoint is called, which I don't think it's a desirable behavior. I'm going to fix that.

I'm glad we had this conversation :smile:

db/migrate/20220205193302_change_notifications.rb
RubenSousaDinis: Ok, all good then!

verdefred (approved):
Looks good 🚀

pdfrod:
@francisco-leal let me know when this can be merged to `dev` or feel free to do it yourself.

app/notifications/base_notification.rb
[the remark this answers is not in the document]
RubenSousaDinis: Got it!

app/services/create_notification.rb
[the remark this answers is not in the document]
RubenSousaDinis: Ok, let's go with the scope only then

francisco-leal:
I'll rebase it to resolve conflicts and merge it today

francisco-leal (approved):

PR #328, opened 2022-02-25
Create public profile

verdefred:
## Summary
Create a public lite profile page

### What changed?
A new navbar for not logged in users
Talent profile accessible by anyone
Hide all tabs but "overview"
Create a new token details component
Remove actionable buttons for public page

## Notion link
https://www.notion.so/talentprotocol/Public-Profile-lite-version-2-2-45183d7094004beab55852fe3e3b16e6

## How to test?
Open a talent's page while logged in and while logged out.
Check if buttons and tabs appear when logged in and don't appear when logged out

### DIFF
--- a/app/packs/src/components/talent/Show/SimpleTokenDetails.jsx
+++ b/app/packs/src/components/talent/Show/SimpleTokenDetails.jsx
@@ -1,0 +1,67 @@
+import React, { useState, useEffect } from "react";
+import { ethers } from "ethers";
+import currency from "currency.js";
+
+import { H4, P2 } from "src/components/design_system/typography";
+
+import {
+  ApolloProvider,
+  useQuery,
+  GET_TALENT_PORTFOLIO_FOR_ID_SIMPLE,
+  client,
+} from "src/utils/thegraph";
+
+const SimpleTokenDetails = ({ token, ticker }) => {
+  const { loading, data } = useQuery(GET_TALENT_PORTFOLIO_FOR_ID_SIMPLE, {
+    variables: { id: token?.contract_id?.toLowerCase() },
+  });
+
+  const [tokenData, setTokenData] = useState({
+    price: 0.1,
+    totalSupply: 0,
+    supporterCount: 0,
+  });
+
+  useEffect(() => {
+    if (loading || !data || !data.talentToken) {
+      return;
+    }
+
+    setTokenData({
+      ...tokenData,
+      totalSupply: ethers.utils.formatUnits(data.talentToken.totalSupply || 0),
+      supporterCount: data.talentToken.supporterCounter || 0,
+    });
+  }, [data, loading]);
+
+  const formatNumberWithSymbol = (value) => currency(value).format();
+
+  const formatNumberWithoutSymbol = (value) =>
+    currency(value, { symbol: "" }).format();
+
+  return (
+    <div>
+      <div className="card disabled d-flex flex-column align-items-center justify-content-center mb-5">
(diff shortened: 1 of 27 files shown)


app/views/talent/show.html.erb
```diff
@@ -11,9 +17,9 @@
         careerGoal: @talent.career_goal.as_json,
         goals: @talent.career_goal.present? ? @talent.career_goal&.goals.order(:due_date) : [],
         milestones: @talent.milestones.as_json,
-        currentUserId: current_user.id,
+        currentUserId: current_user&.id,
```
RubenSousaDinis, line 20: Not related with the changes, but would it be too difficult to start using the new serializer here?

verdefred: Yup, makes sense! I'll make that change as well

app/packs/src/components/talent/Show/SimpleTokenDetails.jsx
```diff
@@ -0,0 +1,67 @@
...
+    });
+  }, [data, loading]);
+
+  const formatNumberWithSymbol = (value) => currency(value).format();
+
+  const formatNumberWithoutSymbol = (value) =>
+    currency(value, { symbol: "" }).format();
+
+  return (
+    <div>
```
francisco-leal, line 43: do we need the extra div here?

app/packs/src/components/talent/TalentShow.jsx
```diff
@@ -121,16 +119,18 @@ const TalentShow = ({
...
+    if (searchParams.get("tab") && !publicPageViewer) {
       setPageInDisplay(searchParams.get("tab"));
-    } else {
+    } else if (!publicPageViewer) {
       window.history.pushState(
         {},
         document.title,
         `${url.pathname}?tab=overview`
       );
+    } else if (publicPageViewer) {
```
francisco-leal, line 130: since publicPageViewer is a boolean this can be an else statement right?

app/packs/src/components/top_bar/LoggedOutTopBar.jsx
```diff
@@ -0,0 +1,65 @@
...
+              ) : (
+                <LogoDark width={128} height={20} />
+              )}
+            </a>
+          )}
+        </div>
+        <div className="d-flex" style={{ height: 34 }}>
+          <Button
+            type="primary-default"
+            onClick={() => window.open(TALENT_PROTOCOL_WEBSITE)}
```
francisco-leal, line 33: I'm not sure if this should link to the website or directly to the typeform, can you check with Marketing?

francisco-leal (commented):
Just some minor details

app/packs/src/components/talent/Show/SimpleTokenDetails.jsx
verdefred: we don't, will delete. nice catch

app/packs/src/components/talent/TalentShow.jsx
verdefred: it can be an else, correct. I'll make the change

app/packs/src/components/top_bar/LoggedOutTopBar.jsx
verdefred: Talked with marketing and there is a way to directly send the user to the form and that's the preferred link here. I'll make that change, thanks!

francisco-leal (approved):

PR #336, opened 2022-03-01
Notification emails

francisco-leal:
## Summary of changes

A notifications preference json field was added to the users model. This allows the user to choose one of three things, if they want to receive email notifications immediately, as a digest or don't receive them at all.

### DIFF
--- a/app/jobs/send_notification_digests_job.rb
+++ b/app/jobs/send_notification_digests_job.rb
@@ -1,0 +1,14 @@
+class SendNotificationDigestsJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    notifications = Notification.where(read_at: nil, emailed_at: nil)
+      .where("created_at >= ?", 2.day.ago)
+      .pluck(:recipient_id, :id).group_by(&:first)
+      .transform_values { |pairs| pairs.map(&:second) }
+
+    notifications.each do |user_id, notification_ids|
+      NotificationMailer.digest(user_id, notification_ids).deliver_later
+    end
+  end
+end
--- a/app/jobs/send_notification_digests_job.rb
+++ b/app/jobs/send_notification_digests_job.rb
@@ -2,13 +2,17 @@ class SendNotificationDigestsJob < ApplicationJob
   queue_as :default
 
   def perform
-    notifications = Notification.where(read_at: nil, emailed_at: nil)
-      .where("created_at >= ?", 2.day.ago)
-      .pluck(:recipient_id, :id).group_by(&:first)
-      .transform_values { |pairs| pairs.map(&:second) }
+    notifications = Notification
+      .where(read_at: nil, emailed_at: nil)
+      .where("created_at >= ?", 2.days.ago)
+      .group(:recipient_id)
+      .select(
+        "recipient_id",
+        'array_agg("id") as notification_ids'
+      )
 
-    notifications.each do |user_id, notification_ids|
-      NotificationMailer.digest(user_id, notification_ids).deliver_later
+    notifications.each do |row|
+      NotificationMailer.digest(row.recipient_id, row.notification_ids).deliver_later
     end
   end
 end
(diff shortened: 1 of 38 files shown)


app/jobs/send_notification_digests_job.rb
```diff
@@ -0,0 +1,14 @@
+class SendNotificationDigestsJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    notifications = Notification.where(read_at: nil, emailed_at: nil)
```
RubenSousaDinis, line 5: In case you want to use only a Database query and prevent memory computations:

```
query = <<~SQL
  select n.recipient_id, array_agg(n.id) as notification_ids
  from notifications n
  where n.read_at IS NULL
  AND n.emailed_at IS NULL
  AND n.created_at >= :created_date
  GROUP BY n.recipient_id
SQL

query_params = {
  created_date: Date.today - 2.days
}

sql = ActiveRecord::Base.sanitize_sql_array([
  query,
  query_params
])

ActiveRecord::Base.connection.execute(sql).to_a
```

francisco-leal: I don't love the readability of pure SQL 😅  but it will be more performant for sure. if you and @verdefred are ok with it we can go with the more performant solution

RubenSousaDinis: Ahah Active Record is just a way to achieve pure SQL and I think this SQL is more explicit than the existing solution. Probably the more complicated part is the `array_agg` function. 

I'm ok if you go with your solution tho. It's running in an async job.

config/sidekiq.yml
```diff
@@ -6,6 +6,9 @@
   - mailers
   - low
 :schedule:
+  send_notification_digests:
+    cron: '0 1 * * *'
```
RubenSousaDinis, line 10: There're some studies that point 10am to be the best time to send emails in order to get more openings and clicks. Not sure if that's what we want here since it's a digest but thought it might be worth flagging.

app/jobs/send_notification_digests_job.rb
francisco-leal: I've used array_agg in the past, it's really helpful! I'm fine with it, just want to get Fred's opinion as well

app/jobs/send_notification_digests_job.rb
```diff
@@ -0,0 +1,14 @@
...
+      .where("created_at >= ?", 2.day.ago)
```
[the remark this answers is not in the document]
francisco-leal: It's just so if someone has disabled the notifications and then turns them on they won't receive everything that is on the "backlog"

app/packs/src/components/design_system/link/index.jsx
```diff
@@ -15,11 +15,21 @@ const Link = ({
   target,
   className,
   children,
+  onClick,
```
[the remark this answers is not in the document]
francisco-leal: I guess maybe the best approach, picking up on what you suggested is to if we get an href, then render an anchor and if we get a onClick then we render a button, what do you think? but always in the same component

config/sidekiq.yml
francisco-leal: Currently most of our userbase is european, eastern Europe I believe. I'll set it to our 10am and we can update it in the future

app/jobs/send_notification_digests_job.rb
francisco-leal: Going to make this change then 👍

francisco-leal: I've used a closer approach to yours but still keep it in ActiveRecord, what do you think @RubenSousaDinis?

app/packs/src/components/design_system/link/index.jsx
```diff
@@ -14,11 +14,32 @@ const Link = ({
...
+      </button>
+    );
+  }
+
   return (
     <a
       className={cx("link-container", type, disabled && "disabled", className)}
-      href={disabled ? null : href}
+      href={href}
+      disabled={disabled}
```
verdefred, line 42: anchor doesn't have a disable prop so here we need to do
`href={disabled ? null : href}` instead

verdefred (commented):
great work, just left a small comment

verdefred (approved):

PR #337, opened 2022-03-01
Small tweaks to messages UI

francisco-leal:
## Summary of changes

* When sending a message to a new user it was showing up as it being sent 21 years ago, this was due to applying a format on the date twice.
* We already have a lot of users in production and so when opening the new messages modal it was trying to list 15k users which timeouts on the server as we're trying to collect a LOT of profile picture URLs, so only when you input some something on the name will it actually search

### DIFF
--- a/app/packs/src/components/chat/MessageUserList.jsx
+++ b/app/packs/src/components/chat/MessageUserList.jsx
@@ -27,6 +27,7 @@ const UserMessage = ({ user, activeUserId, onClick, mode }) => {
 
   const displayTime = () => {
     if (user.last_message_date) {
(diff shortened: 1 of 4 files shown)


app/packs/src/components/chat/MessageUserList.jsx
```diff
@@ -27,6 +27,7 @@ const UserMessage = ({ user, activeUserId, onClick, mode }) => {
 
   const displayTime = () => {
     if (user.last_message_date) {
+      console.log(user.last_message_date);
```
verdefred, line 30: extra console.log

PR #344, opened 2022-03-05
Display links in messages

RubenSousaDinis:
## Summary

Screenshot 2022-03-07 at 09 12 33

Screenshot 2022-03-14 at 20 13 37

### What changed?

It's now possible to have clickable links in the chat.

### Why it changed?

Adds more usability to the chat. No need to copy and paste urls.

### How it changed?

Added a lib (https://github.com/anantoghosh/react-linkify-it) to detect urls and make them clickable.

## Notion link

https://www.notion.so/Prevent-phising-links-on-chat-a1e9488901514ce193b25776bd3535b9

## How to test?

Send links, twitter handles and emails and check they open correctly.

## Notes

We might want to add whitelisted URLs in the future. Right now the user needs to confirm  link every time he clicks it.

### DIFF
--- a/app/packs/src/components/chat/Message.jsx
+++ b/app/packs/src/components/chat/Message.jsx
@@ -1,9 +1,20 @@
 import React from "react";
 import dayjs from "dayjs";
+import { LinkItEmail, LinkItUrl } from 'react-linkify-it';
 
 import TalentProfilePicture from "../talent/TalentProfilePicture";
 import P2 from "src/components/design_system/typography/p2";
 import P3 from "src/components/design_system/typography/p3";
+
+const MessageLinkified = (props) => {
+  return (
+    <LinkItUrl>
+      <LinkItEmail>
+          {props.message}
+      </LinkItEmail>
+    </LinkItUrl>
+  )
+}
 
 const Message = (props) => {
   const {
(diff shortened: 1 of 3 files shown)


app/packs/src/components/chat/Message.jsx
```diff
@@ -1,10 +1,23 @@
...
 
 import TalentProfilePicture from "../talent/TalentProfilePicture";
 import P2 from "src/components/design_system/typography/p2";
 import P3 from "src/components/design_system/typography/p3";
 
+const MessageLinkified = (props) => {
+  return (
+    <LinkItUrl>
+      <LinkItEmail>
+        <LinkItTwitter>
```
verdefred, line 13: not sure if we should have this one. I can maybe want to pass my telegram handle or something else and it'll link to a random twitter account. what do you think, @RubenSousaDinis @francisco-leal?

francisco-leal: Agreed, ideally, it should even be their own talent protocol account

RubenSousaDinis: Good point! Removed

verdefred (approved):

RubenSousaDinis:
> 



> > @RubenSousaDinis this PR doesn't include a confirmation modal to guarantee that the user confirms he wants to visit that url?
> 
> @francisco-leal I was thinking on adding that on another PR, since this one already delivers value. Should I add it to this one?

> @francisco-leal added to this PR.

francisco-leal (approved):

verdefred (approved):

PR #359, opened 2022-03-09
Add tag with link to discovery row

francisco-leal:
## Summary

For future partnerships, we want to showcase talent on the app that belongs to those partnerships. As such we need to add a tag next to the title and it needs to include a link to send to that partners page

### DIFF
--- a/app/packs/src/components/discovery/discovery_rows.jsx
+++ b/app/packs/src/components/discovery/discovery_rows.jsx
@@ -1,9 +1,10 @@
 import React, { useState, useCallback } from "react";
 import { useWindowDimensionsHook } from "src/utils/window";
-import { P1 } from "src/components/design_system/typography";
+import { P1, P3 } from "src/components/design_system/typography";
 import { Caret } from "src/components/icons";
 import NewTalentCard from "src/components/design_system/cards/NewTalentCard";
 import Button from "src/components/design_system/button";
+import Tag from "src/components/design_system/tag";
 
 import cx from "classnames";
 
@@ -63,6 +64,25 @@ const DiscoveryRows = ({ discoveryRows, updateFollow }) => {
   const disableRight = (title, talents) =>
     start[title] + itemsPerRow(talents) >= talents.length;
 
+  const discoveryTag = (row) => {
+    console.log(row);
+    if (!!row.tag_link) {
+      return (
+        <a href={row.tag_link} className="cursor-pointer ml-2 " target="_blank">
+          <Tag className="bg-primary text-white cursor-pointer">
+            <P3 className="current-color" bold text={row.tag} />
+          </Tag>
+        </a>
+      );
+    } else {
+      return (
+        <Tag className="ml-2 bg-primary text-white">
+          <P3 className="current-color" bold text={row.tag} />
+        </Tag>
+      );
+    }
+  };
+
   return (
     <>
       {discoveryRows.map((row) => (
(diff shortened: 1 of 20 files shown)


app/packs/stylesheets/_types.scss
```diff
@@ -66,6 +66,7 @@
   font-weight: 400;
   line-height: 19.8px;
   margin: 0;
+  letter-spacing: 0.02rem;
```
verdefred, line 69: good catch, can you add this to .p2 as well, please?

app/packs/stylesheets/custom/_shared.scss
```diff
@@ -113,6 +113,10 @@
   color: $white !important;
 }
 
+.dark-body .text-white {
+  color: $black !important;
+}
+
```
verdefred, line 119: I think this will break a lot of things. we use text-white when we always want white text, independent of the mode. maybe we need better naming for this

db/migrate/20220309094506_add_tag_to_discovery_row.rb
```diff
@@ -0,0 +1,5 @@
+class AddTagToDiscoveryRow < ActiveRecord::Migration[6.1]
+  def change
+    add_column :discovery_rows, :tag, :string
```
verdefred, line 3: I think the naming is a bit weird, since we already have associated tags on discovery row. it will not be easy, in the future, to understand the different meanings, imo. if we don't want to use `partnership` because it's too specific, maybe `subtitle` or `placeholder`?

lib/tasks/tmp/add_impact_market_row.rake
```diff
@@ -0,0 +1,18 @@
...
+      user = User.find_by(username: username)
+      TalentTag.create(talent: user.talent, tag: tag)
+    end
+
+    row = DiscoveryRow.create!(
+      title: "Women of Impact",
+      tag: "IMPACT MARKET",
+      tag_link: "https://www.impactmarket.com/"
+    )
+    tag.discovery_row_id = row.id
```
verdefred, line 16: I think we're missing a save after this

lib/tasks/tmp/add_impact_market_row.rake
```diff
@@ -0,0 +1,18 @@
...
+    tag = Tag.create(description: "impact_market_hidden", hidden: true)
+    usernames = ["jessicagbt", "rosaline2104", "anaccmn", "avina", "susannezapelao", "hannahbamwerinde"]
+
+    usernames.each do |username|
+      user = User.find_by(username: username)
+      TalentTag.create(talent: user.talent, tag: tag)
+    end
+
+    row = DiscoveryRow.create!(
+      title: "Women of Impact",
```
verdefred, line 12: shouldn't the title be "Women of Impact Market"? I think that's what I saw somewhere in our product channel

app/packs/src/components/discovery/discovery_rows.jsx
```diff
@@ -63,18 +64,41 @@ const DiscoveryRows = ({ discoveryRows, updateFollow }) => {
   const disableRight = (title, talents) =>
     start[title] + itemsPerRow(talents) >= talents.length;
 
+  const discoveryTag = (row) => {
+    console.log(row);
```
verdefred, line 68: extra console.log

lib/tasks/tmp/add_impact_market_row.rake
francisco-leal: No, since we're having the tag "IMPACT MARKET" next to it it would be to redudant

app/packs/stylesheets/custom/_shared.scss
francisco-leal: We're already doing the same thing with text-black, which was the reason for this approach

francisco-leal: `text-white` should work the same as `text-black`, if we want to have a specific class that doesn't change based on the mode than it should be something different

lib/tasks/tmp/add_impact_market_row.rake
verdefred: That leads me to my second question, then. Should the title only be "Women of" since the tag already has "IMPACT" on it?

francisco-leal: It doesn't read as nice as you think 😅

verdefred (dismissed):

PR #362, opened 2022-03-09
Create Supporter Profile page

verdefred:
## Summary
Edit profile for supporters now with an About section
Supporters can create a display name, occupation, tags, website, socials and location
Supporter profile page now live both public and private

## Notion link
https://www.notion.so/talentprotocol/Supporter-Profile-2-2-562efec5abc843abb08935fa521e882f

## How to test?
Check if current edit profile for talents keeps working as before
Check if new edit profile for supporters is working correctly
As a supporter, try to open your own profile and check if everything is working correctly


app/controllers/users_controller.rb
```diff
@@ -20,12 +20,16 @@ def index
...
-    )
+    if talent
+      @talent = TalentBlueprint.render_as_json(
+        talent,
+        view: :extended,
+        current_user: current_user,
+        tags: @user.tags.where(hidden: false)
+      )
+    else
+      puts "investor"
```
verdefred, line 31: TODO: next PR with show for supporters

config/routes.rb
```diff
@@ -87,6 +82,18 @@
 
   resources :wait_list, only: [:create, :index]
 
+  namespace :api, defaults: {format: :json} do
+    namespace :v1 do
+      resources :talent, only: [:show]
+    end
+  end
```
verdefred, line 89: not sure if this is correct here or if it should be moved to a different controller altogether, so everything inside api/v1 is for users logged in only. let me know what you think @francisco-leal @RubenSousaDinis

app/packs/src/components/supporters/edit/Profile.jsx
```diff
@@ -0,0 +1,168 @@
...
+  const { mobile } = useWindowDimensionsHook();
+  const [tabHasChanges, setTabHasChanges] = useState(false);
+  const [show, setShow] = useState(false);
+  const [nextTab, setNextTab] = useState(null);
+  const [sharedState, setSharedState] = useState({
+    ...props,
+    currentPassword: "",
+    newPassword: "",
+    deletePassword: "",
+    railsContext: null,
```
francisco-leal, line 32: why do we need to force the rails context to be null on the shared state?

app/views/users/show.html.erb
```diff
@@ -1,10 +1,21 @@
...
+  <% end %>
+<% else %>
+  <% content_for :meta_title, "#{@investor["user"]["username"]} | #{DEFAULT_META["meta_product_name"]}" %>
+  <% if @investor["headline"].present? %>
+    <% content_for :meta_description, @investor["headline"] %>
+  <% end %>
+  <% if @investor["profile_picture_url"].present? %>
+    <% content_for :meta_image, @investor["profile_picture_url"] %>
+    <% content_for :meta_card_type, "summary" %>
+  <% end %>
```
francisco-leal, line 18: Can we do this based on the @user instead of talent vs investor? the attributes are the same as well so I don't think we need these many if/else

config/routes.rb
francisco-leal: What data are you requesting async from the backend here? the profile pictures?

francisco-leal (dismissed):
Just left some comments but overall nothing stands out as requiring code changes. Good job! 💪

app/packs/src/components/supporters/edit/Profile.jsx
verdefred: It's not needed, it's just to not send it to the backend since we're sending the whole shared state

app/views/users/show.html.erb
verdefred: good idea, I'll do that 👍

verdefred:
https://github.com/talentprotocol/mvp/pull/362#discussion_r825825125
Not anymore. Now we need more info, because of the Supporting table in the Supporter Profile. We're requesting the normal view of TalentBlueprint

francisco-leal (approved):

PR #374, opened 2022-03-15
Add lambda integration

francisco-leal:


lib/tasks/airdrop_level_one.rake
```diff
@@ -2,14 +2,19 @@ require "csv"
 
 namespace :users do
   task airdrop_level_one: :environment do
-    puts "Sending to ##{User.count} users"
+    index = 0
+    total = User.count
```
verdefred, line 6: total = User.where.not(wallet_id: nil).count

PR #402, opened 2022-03-18
Add pagination

francisco-leal:
## Summary

We need to reduce the number of requests we do to the backend so that we can have DDoS prevention. So we want to batch the number of requests we do to get data to fill out tables

### DIFF
--- a/app/controllers/api/v1/supporters_controller.rb
+++ b/app/controllers/api/v1/supporters_controller.rb
@@ -1,0 +1,23 @@
+class API::V1::SupportersController < ApplicationController
+  # This is a public endpoint
+  def index
+    @users = User.includes(:talent, :investor).where(wallet_id: wallet_ids)
+    user_data = @users.map do |u|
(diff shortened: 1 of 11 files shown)


app/controllers/api/v1/supporters_controller.rb
```diff
@@ -0,0 +1,23 @@
+class API::V1::SupportersController < ApplicationController
+  # This is a public endpoint
+  def index
+    @users = User.includes(:talent, :investor).where(wallet_id: wallet_ids)
+    user_data = @users.map do |u|
+      {
+        id: u.id,
+        wallet_id: u.wallet_id,
+        profilePictureUrl: u&.talent&.profile_picture_url || u.investor.profile_picture_url,
```
verdefred, line 9: I think if a talent exists but doesn't have a profile picture this will break because the investor will not exist

verdefred (approved):

PR #420, opened 2022-03-24
Add member nfts

francisco-leal:
## Summary

Adds a service to generate the dynamic member NFT image and airdrop to a user


verdefred (dismissed):

app/jobs/send_member_nft_to_user_job.rb
```diff
@@ -0,0 +1,14 @@
+class SendMemberNFTToUserJob < ApplicationJob
+  queue_as :default
+
+  def perform(user_id:)
+    if ENV["ENABLE_MEMBER_USER_NFT"] != "enable"
```
RubenSousaDinis, line 5: one line?

db/migrate/20220323112924_add_member_nft_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddMemberNFTToUsers < ActiveRecord::Migration[6.1]
+  def change
```
RubenSousaDinis, line 2: I was taking a look into the user table and we already have 4 fields representing `user_nft` data. We currently have 28 columns in the table. IMO we should start paying more attention to this and start splitting this data. Should we have a `user_nfts` table with an enum that tells if the nft is a user or member nft?

app/services/NFT/generate_member_nft.rb
```diff
@@ -0,0 +1,49 @@
+module NFT
+  class GenerateMemberNFT
+    def initialize(user:)
+      @user = user
+      @image_path = ENV["ENABLE_COMMUNITY_MEMBER_NFT"] == "enable" ? "public/Member_NFT02.jpg" : "public/Member_NFT01.jpg"
```
RubenSousaDinis, line 5: Picky request.

```suggestion
      @image_path = ENV["ENABLE_COMMUNITY_MEMBER_NFT"] == "enable" ? "public/member_nft_02.jpg" : "public/member_nft_01.jpg"
```

db/migrate/20220323112924_add_member_nft_to_users.rb
francisco-leal: Yes, but I already have that ready in a separate branch that introduces the Perks as NFTs, and it will be easier to not have conflicts like this. Will be fixed very soon!

RubenSousaDinis (dismissed):

verdefred (dismissed):

verdefred (approved):

PR #437, opened 2022-04-06
Setup invite races

francisco-leal:
## Summary

We want an automated leaderboard with rewards distribution embedded into the application

### What changed?

Adds a few things:
* Adds a Reward model to be able to assign any reward to any user, the valid categories are: `"QUEST", "RACE", "TALENT_INVITE", "OTHER"`
* Adds a Race model to track every race and to associate a race with users upon creation. We use the race_id on each user to calculate who was registered to the platform. This is better than using new invites each week since we'll make sure the user is still part of the platform when we distribute rewards (preventing someone from deleting the account and re-registering)
* Adds a new Earn screen to showcase race leaderboards and the number of invites you have

### Why it changed?

We need to push for growth and gamifiying this experience and showing the results is the best way to do it currently


db/migrate/20220330143340_create_races.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRaces < ActiveRecord::Migration[6.1]
+  def change
+    create_table :races do |t|
+      t.date :started_at
```
RubenSousaDinis, line 4: should we also store minutes? Might be useful

db/migrate/20220330145115_create_rewards.rb
```diff
@@ -1,9 +1,9 @@
 class CreateRewards < ActiveRecord::Migration[6.1]
   def change
     create_table :rewards do |t|
-      t.integer :required_amount
-      t.text :description
-      t.references :talent, null: false, foreign_key: true
+      t.references :user, null: false, foreign_key: true
+      t.string :reason
+      t.integer :amount
```
RubenSousaDinis, line 7: might be safe to use bigint. Although I'm almost sure we will not give that much 😄

app/controllers/rewards_controller.rb
```diff
@@ -0,0 +1,20 @@
+class RewardsController < ApplicationController
+  def index
+    invites = current_user.invites
+    @talent_invites = invites.select { |i| i.talent_invite }
```
RubenSousaDinis, line 4: we can't use active record here?

app/controllers/rewards_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    @supporter_invites = invites.select { |i| !i.talent_invite }
+
+    @race = Race.active_race || Race.last
+    @rewards = current_user.rewards
+
+    if @talent_invites.length > 0
+      invite_ids = @talent_invites.map(&:id)
+      talents = Talent.joins(:user).where(user: {invite_id: invite_ids})
```
RubenSousaDinis, line 12: You can use `talents = Talent.joins(:user).where(user: {invite_id: @talent_invites})` if @talent_invites is an ActiveRecord Relation

app/models/reward.rb
```diff
@@ -0,0 +1,14 @@
+class Reward < ApplicationRecord
+  belongs_to :user
+  validate :category_is_valid
+
+  CATEGORIES = ["QUEST", "RACE", "TALENT_INVITE", "OTHER"]
```
RubenSousaDinis, line 5: This works. But just wondering why we're not using an enum. Validation will be there by default and it also give us filtering capabilities.

db/migrate/20220404074010_add_race_id_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddRaceIdToUser < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :users, :race
```
RubenSousaDinis, line 3: This works but will not give you an idea on how much races the user participated in the past. We will need to replace the `race_id` every time a new race is created. Can we have several races at the same time? I would create a new relation table `user_races`.

app/controllers/rewards_controller.rb
francisco-leal: We can, but they will never be a lot of invites (they should be 1 of each) and so doing two queries is slower than splitting them in memory

db/migrate/20220330145115_create_rewards.rb
francisco-leal: Agreed, will change that

db/migrate/20220330143340_create_races.rb
francisco-leal: Sure, I'll make that change

app/models/reward.rb
francisco-leal: I don't like ENUMS I like reading the row and being able to understand right away what the value should be

db/migrate/20220404074010_add_race_id_to_user.rb
francisco-leal: That is not how the `race_id` works. We assign a race when the user registers and never again. As a user you can invite others in any race and when another user joins through your link he will get the race_id populated

RubenSousaDinis: Got it!

app/services/finish_active_race.rb
```diff
@@ -0,0 +1,36 @@
...
+    race = Race.new
+    start_date = (last_race.ends_at + 1.day).beginning_of_day
+    race.started_at = start_date.beginning_of_day
+    race.ends_at = start_date + 1.week - 1.second
+    race.save!
+  end
+
+  # Reward the winners of the race that just finished
+  def reward_winners
+    winners = @race.results.limit(3).to_a
```
verdefred, line 26: should we order the results to make sure the order of the winners is correct?

app/services/prepare_race_results.rb
```diff
@@ -0,0 +1,45 @@
+class PrepareRaceResults
+  def initialize(race:, user:)
+    @race = race
+    @user = user
+  end
+
+  def call
+    results = @race.results.to_a
```
verdefred, line 8: should this results also have an order to make sure the position is correct?

app/packs/stylesheets/custom/_rewards.scss
```diff
@@ -0,0 +1,109 @@
...
+  .race-time-counter-box {
+    width: 80px;
+    height: 80px;
+  }
+
+  .talent-rewards-header-row {
+    flex-direction: column;
+    justify-content: start;
+  }
+}
```
verdefred, line 109: end of file new line missing

app/packs/src/components/rewards/ReferralRace.jsx
```diff
@@ -0,0 +1,483 @@
...
+    days: 7,
+    hours: 0,
+    minutes: 0,
+    seconds: 0,
+  });
+  let timeoutPointer;
+
+  const updateTimeUntilEnd = () => {
+    const currentTime = dayjs();
+    let calculatedTime = dayjs(race.ends_at).endOf("day");
```
verdefred, line 33: I think we can improve a little bit this name. something like `timeWhenRaceEnds`?

app/services/finish_active_race.rb
francisco-leal: The query already has an order clause

app/services/prepare_race_results.rb
francisco-leal: The query already has an order clause

verdefred (approved):

RubenSousaDinis (approved):

PR #442, opened 2022-04-08
Add boolean to store if message was sent to all supporters

francisco-leal:
## Summary

We should store if a message was sent using the new "Send Message to All supporters" feature so that in the future we can use that flag to change the UI


RubenSousaDinis (approved):
Well done!

PR #451, opened 2022-04-14
Improve product onboarding

verdefred:
## Summary
We want to improve the product onboarding so it's easier and more gamified. This way we'll hopefully have more people completing their profile and interacting with the platform.
For this we decided to create a Quests page, where we have different challenges and rewards.
For this to be possible we needed to add:
- Quest model
- Task model (tasks inside a quest with something to do and a reward)
- ProfilePageVisitor to track different IPs accessing profile pages (since IP is PII we needed to encrypt this column)
- Notification once a quest is completed
- Rake task to populate previous users' quests
- Rake task to set the correct status of previous users' quests
- Quests page
- Tasks page

## Notion link
https://www.notion.so/talentprotocol/Improve-Product-Onboarding-2-2-1fd900f6e3454a29a76aef3ddb81bcfd

## How to test?
Go to the earn page and try to complete a few quests


app/services/create_user.rb
```diff
@@ -108,4 +110,24 @@ def create_invite(user)
...
+  def create_quests(user)
+    Quests::PopulateForUser.new.call(user: user)
+  end
+
+  def give_reward_to_inviter(invite)
+    return unless invite.user
+
+    task_done = Task
+      .joins(:quest)
+      .where(title: "Get 5 people to register")
```
RubenSousaDinis, line 123: We're using this title strings several times in different parts of the app. If for some reason we want to change the wording we need to change it those places. There's a file change in the locales file. Should we add these strings there?

db/migrate/20220412143224_create_profile_page_visitors.rb
```diff
@@ -0,0 +1,11 @@
+class CreateProfilePageVisitors < ActiveRecord::Migration[6.1]
+  def change
+    create_table :profile_page_visitors do |t|
+      t.text :ip_ciphertext, null: false
+      t.string :ip_bidx, index: true
+      t.references :user, null: false, foreign_key: true, index: true
+
```
RubenSousaDinis, line 7: Should we add a `last_visited_at` timestamp? Might be useful

app/controllers/api/v1/follows_controller.rb
```diff
@@ -14,6 +14,18 @@ def create
...
+        .joins(:quest)
+        .where(title: "Add 3 talent to watchlist")
+        .where(quest: {user: current_user})
+        .take
+        .done?
+
+      if Follow.where(follower_id: current_user.id).count > 2 && !task_done
+        current_user.invites.where(talent_invite: false).update_all(max_uses: nil)
+        Quests::Update.new.call(title: "Add 3 talent to watchlist", user: current_user)
+      end
```
francisco-leal, line 27: If we're not returning anything to the frontend I would push this type of calculation to sidekiq so that if there is an error the follow still comes through and to reduce the number of calculations we need to do on the controller

db/migrate/20220412143224_create_profile_page_visitors.rb
francisco-leal: If we're not creating multiple entries per user then yes, otherwise it will just be the creation date

francisco-leal (approved):

PR #460, opened 2022-04-21
Add links to tasks

verdefred:
## Summary
- Move description and rewards to FE so we can have links on them
- Fix some css issues with cards
-  Add STI to Quest model
- Fix copy
- Add missing components

## Notion link

https://www.notion.so/talentprotocol/Improve-Product-Onboarding-2-2-1fd900f6e3454a29a76aef3ddb81bcfd


app/models/quests/scout.rb
```diff
@@ -0,0 +1,11 @@
+module Quests
```
RubenSousaDinis, line 1: Let's start adding `# frozen_string_literal: true` comment to files with string literals. Right now every time a file with string literals are initialised we need to allocate more memory to store the string literals on the file. Let's imagine this file in initialised 1000 times. We will be allocating 1000x two string literals in memory. By freezing string literals we will only use one space in memory per string literal.

There're conversations to make this the default behaviour in ruby 3, but we're running an older version. I believe we can still make this behaviour a default in our ruby version instead of adding the magic comment on each file.

francisco-leal (approved):

PR #461, opened 2022-04-21
Add chat model

RubenSousaDinis:
## Summary

We're struggling to paginate chats. Currently we're sorting and searching in the client and we should start doing that in the server in order to add pagination.

This PR adds a proposal for a chat model that will store high level information about a chat between two users, such as:

- the last time a message was sent between them (useful for sorting)
- the content of the last message sent


lopeselio:
@RubenSousaDinis Are the messages sent in the order of the most recent messages according to the date of the message sent, in this chat model? I am starting to work on the pagination aspect.

RubenSousaDinis:
> @RubenSousaDinis Are the messages sent in the order of the most recent messages according to the date of the message sent, in this chat model? I am starting to work on the pagination aspect.

@lopeselio yes. They will come order. If they are not already I will do that.

francisco-leal (approved):

PR #478, opened 2022-05-02
Change races conditions to only count users that buy talent tokens

verdefred:
## Summary
We want to improve our metrics concerning active users. We consider an active user an user that buys talent tokens.
For this to happen we'll be changing the referral race to only allow users that buy talent tokens

## How to test?
Open the referral race page and check the results


app/services/races/results_one_and_two_sql.rb
```diff
@@ -0,0 +1,17 @@
+module Races
+  class ResultsOneAndTwoSql
```
francisco-leal, line 2: I think a better name could be `ResultsWithBasicInvites`

app/services/races/results_three_and_four_sql.rb
```diff
@@ -0,0 +1,19 @@
+module Races
+  class ResultsThreeAndFourSql
```
francisco-leal, line 2: `ResultsWithWalletConnected`

francisco-leal (approved):

francisco-leal (approved):

PR #488, opened 2022-05-04
Talent quest goes to form for supporters

verdefred:
## Summary
We came to the conclusion that it didn't make sense for a supporter to make the tasks on the Talent quest, so now the user is redirected to our form to apply for launching a token

This PR also fixes a concurrency issue with the CreateProfilePageVisitor

### DIFF
--- a/app/services/api/create_profile_page_visitor.rb
+++ b/app/services/api/create_profile_page_visitor.rb
@@ -1,9 +1,16 @@
 module API
   class CreateProfilePageVisitor
-    def call(ip:, user:)
-      profile_page_visitor = ProfilePageVisitor.find_or_create_by!(user: user, ip: ip)
-      profile_page_visitor.update(last_visited_at: Time.current)
+    MAX_RETRIES = 5
 
+    def call(ip:, user:)
+      count = 0
+      begin
+        profile_page_visitor = ProfilePageVisitor.find_or_create_by!(user: user, ip: ip)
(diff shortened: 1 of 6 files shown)


app/services/api/create_profile_page_visitor.rb
```diff
@@ -1,11 +1,19 @@
...
-      profile_page_visitor.update(last_visited_at: Time.current)
+      count = 0
+      begin
+        profile_page_visitor = ProfilePageVisitor.create!(user: user, ip: ip)
+        profile_page_visitor.update(last_visited_at: Time.current)
 
-      if ProfilePageVisitor.where(user: user).count >= (ENV["VISITORS_COUNT"].to_i || 10)
-        UpdateTasksJob.perform_later(type: "Tasks::ShareProfile", user_id: user.id)
+        if ProfilePageVisitor.where(user: user).count >= (ENV["VISITORS_COUNT"].to_i || 10)
+          UpdateTasksJob.perform_later(type: "Tasks::ShareProfile", user_id: user.id)
```
RubenSousaDinis, line 12: might be safer to move this call off of the retry block. With the current implementation is not a problem, but a new dev might think it's ok to move it up when new logic is appended, no?

verdefred: agree, made the change

RubenSousaDinis (approved):

PR #493, opened 2022-05-09
Resend notifications when the last notification sent was more than a week ago

RubenSousaDinis:
## Summary

We are not sending new email notifications if the user already has an unread notification. From now on we will trigger a new email if the last unread notification was sent longer than a week ago.


francisco-leal (approved):
🚀

PR #499, opened 2022-05-13
Earn page layout updates

RubenSousaDinis:
## Summary

This PR aims to change the layout of the earn screen.

image

## Notion link

https://talentprotocol.notion.site/Deprecate-Referral-Race-5e06be125e6c4505ad923832c17f29e2


app/blueprints/reward_blueprint.rb
```diff
@@ -0,0 +1,9 @@
+class RewardBlueprint < Blueprinter::Base
+  fields :id
+
+  view :normal do
+    fields :amount, :created_at
```
francisco-leal, line 5: Should we also add the type here? We already use it in some places to split the rewards in the overview section. That way we could also use the blueprint there.

Also I'm starting to think that we should not have associations in the "normal" view since it will make it so that we most often than not end up loading resources we might not need

francisco-leal (approved):

PR #506, opened 2022-05-22
Add 30 day Market Cap Variance to talents

taru5:
## Summary

This PR adds the ability to track the total supply of talent tokens over time to help evaluate the Talent Token’s evolution over time.
It stores the total supply of the token for each day in the subgraph implementation and uses the stored values for the last 30 days to calculate the 30-day Market Cap Variance of talent tokens.
The 30-day Market Cap Variance is displayed in the talents table. 

## Notion link

https://talentprotocol.notion.site/Track-Token-Total-Supply-over-times-7dda974032c8482d9e607ec81158ea15

### DIFF
--- a/app/packs/src/components/talent/TalentPage.jsx
+++ b/app/packs/src/components/talent/TalentPage.jsx
@@ -165,7 +165,8 @@ const TalentPage = ({ talents }) => {
             tokenDayData || [],
             deployDateUTC,
             startDate,
-            endDate
+            endDate,
+            totalSupply
           ),
         };
       }
--- a/app/packs/src/components/talent/TalentPage.jsx
+++ b/app/packs/src/components/talent/TalentPage.jsx
@@ -144,7 +144,7 @@ const TalentPage = ({ talents }) => {
         progress: getProgress(totalSupply, maxSupply),
         marketCap: getMarketCap(totalSupply),
         supporterCounter: getSupporterCount(supporterCounter),
-        marketCapVariance: getMarketCapVariance(tokenDayData),
+        marketCapVariance: getMarketCapVariance(tokenDayData || []),
       })
     );
 
--- a/app/packs/src/components/talent/TalentPage.jsx
+++ b/app/packs/src/components/talent/TalentPage.jsx
@@ -13,6 +13,7 @@ import {
   getSupporterCount,
   getMarketCap,
   getProgress,
+  getMarketCapVariance,
 } from "src/utils/viewHelpers";
 import { compareStrings, compareNumbers } from "src/utils/compareHelpers";
 import { post, destroy } from "src/utils/requests";
@@ -130,25 +131,44 @@ const TalentPage = ({ talents }) => {
     }
 
     const newTalents = data.talentTokens.map(
-      ({ id, totalSupply, maxSupply, supporterCounter, ...rest }) => ({
+      ({
+        id,
+        totalSupply,
+        maxSupply,
+        supporterCounter,
+        tokenDayData,
+        ...rest
+      }) => ({
         ...rest,
         token: { contractId: id },
         progress: getProgress(totalSupply, maxSupply),
         marketCap: getMarketCap(totalSupply),
         supporterCounter: getSupporterCount(supporterCounter),
+        marketCapVariance: getMarketCapVariance(tokenDayData),
       })
     );
 
     setLocalTalents((prev) =>
       Object.values(
         [...prev, ...newTalents].reduce(
-          (result, { id, token, marketCap, supporterCounter, ...rest }) => {
+          (
+            result,
+            {
+              id,
+              token,
+              marketCap,
+              supporterCounter,
+              marketCapVariance,
+              ...rest
+            }
+          ) => {
             result[token.contractId || id] = {
               ...(result[token.contractId || id] || {}),
               id: result[token.contractId || id]?.id || id,
               token: { ...result[token.contractId]?.token, ...token },
               marketCap: marketCap || "-1",
               supporterCounter: supporterCounter || "-1",
+              marketCapVariance: marketCapVariance || "-1",
               ...rest,
             };
 
--- a/app/packs/src/components/talent/TalentPage.jsx
+++ b/app/packs/src/components/talent/TalentPage.jsx
@@ -30,11 +30,29 @@ const TalentPage = ({ talents }) => {
   const theme = useContext(ThemeContext);
   const { mobile } = useWindowDimensionsHook();
   const [localTalents, setLocalTalents] = useState(talents);
+
+  const varianceDays = 30;
+  const msDividend = 1000;
+  const dayInSeconds = 86400;
+  const currentDate = new Date();
+  const endDate =
+    Date.UTC(
+      currentDate.getUTCFullYear(),
+      currentDate.getUTCMonth(),
+      currentDate.getUTCDate(),
+      0,
+      0,
+      0
+    ) / msDividend;
+  const startDate = endDate - varianceDays * dayInSeconds;
+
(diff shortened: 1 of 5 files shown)


app/packs/src/utils/viewHelpers.js
```diff
@@ -36,3 +36,20 @@ export const getProgress = (totalSupply, maxSupply) => {
     return value;
   }
 };
+
+export const getMarketCapVariance = (tokenDayData) => {
```
RubenSousaDinis, line 40: You followed this approach: https://www.investopedia.com/terms/v/volatility.asp#toc-how-to-calculate-volatility, right?

app/packs/src/utils/viewHelpers.js
```diff
@@ -36,3 +36,20 @@ export const getProgress = (totalSupply, maxSupply) => {
...
+  const supplies = tokenDayData.map((data) =>
```
RubenSousaDinis, line 41: should we get only records of the last 30 days?

RubenSousaDinis (commented):
Just some comments and questions. Overall looks good!

app/packs/src/utils/viewHelpers.js
taru5: We require the 30-day Market cap variance, so it fetches the last 30 days data from the graph

RubenSousaDinis: From what I see here: https://github.com/talentprotocol/subgraph/pull/2/files#diff-f54845fd4a2b8020fb39aea2a538a53d6a1791f96eb1d186012a13cd3c73693bR61 we only create the record when a token is minted. This is not the last 30 days. Let's imagine the following scenario:

- I launch my token (2000 TAL are minted). **23-05-2022** supply = 2000
- A supporter buys 1000 TAL of my token 5 days after. **28-05-2022** supply = 3000
- A supporter buys 2000 TAL of my token 31 days after. **23-06-2022** supply = 5000

We will arrive at 23-06-2022 with only 3 records in the graph. The first record from the token launch should not be considered for the variance calculation.

Since the graph is event only we can't ensure we will have a record for each day. So we need to filter the records based on the date.

taru5: Yes, you are right, we will need to compute the day's worth of total supply if the day is not present in the graph.

.github/workflows/talent_ci.yml
```diff
@@ -10,6 +10,11 @@ on:
       - dev
       - master
     types: [ready_for_review, opened, reopened, synchronize]
+  pull_request_target:
```
RubenSousaDinis, line 13: This should not be here.

app/packs/src/components/chat/Message.jsx
```diff
@@ -49,7 +49,7 @@ const MessageLinkified = (props) => {
     }
     e.preventDefault();
     setShowPishingModal(true);
-    setUrl(e.target.origin);
+    setUrl(e.target.href);
```
RubenSousaDinis, line 52: This change should not be here.

app/packs/src/utils/viewHelpers.js
```diff
@@ -36,3 +36,42 @@ export const getProgress = (totalSupply, maxSupply) => {
...
+  const supplies = [];
+  const dayData = {};
+  const dayInSeconds = 86400;
+
+  tokenDayData.forEach((data) => {
+    dayData[data.date] = parseFloat(ethers.utils.formatUnits(data.dailySupply));
+  });
+
+  for (let date = startDate, i = 0; date < endDate; date += dayInSeconds) {
+    if (date < deployDate) {
```
RubenSousaDinis, line 56: Not sure I understood why you need the token deploy date. It will never happen having a record in the tokenDayData that was created before the token was deployed right?

app/packs/src/utils/viewHelpers.js
```diff
@@ -36,3 +36,42 @@ export const getProgress = (totalSupply, maxSupply) => {
...
+  deployDate,
+  startDate,
+  endDate,
+  totalSupply
+) => {
+  const supplies = [];
+  const dayData = {};
+  const dayInSeconds = 86400;
+
+  tokenDayData.forEach((data) => {
```
RubenSousaDinis, line 51: If there're no records in the last 30 days, should we have a guard clause that returns 0 variance?

app/packs/src/utils/viewHelpers.js
```diff
@@ -36,3 +36,42 @@ export const getProgress = (totalSupply, maxSupply) => {
...
+    } else if (dayData[date]) {
+      supplies.push(dayData[date]);
+    } else if (i == 0) {
+      supplies.push(parseFloat(ethers.utils.formatUnits(totalSupply)));
+    } else {
+      supplies.push(supplies[i - 1]);
+    }
+    i++;
+  }
+  const totalDailySupply = supplies.reduce((prev, current) => prev + current, 0);
```
RubenSousaDinis, line 67: in the loop above we are ensuring that we have a value for each day in the last 30 days, right? Do we really need this? Can we not do the exact same calculation using only the days where tokens were staked?

app/packs/src/utils/viewHelpers.js
taru5: For calculating the last 30 variance, we need the last 30 days' daily supply since the token was deployed after the start date, the supply for those days should be considered as zero.

app/packs/src/utils/viewHelpers.js
taru5: If there are no records, we will have to assume the supply stayed constant (as the last supply) for the missing days, to properly calculate the last 30 days variance.

app/packs/src/utils/viewHelpers.js
RubenSousaDinis: So you can pass the token deploy date in the graphQL query if the `start_date < token deploy date`, no?

taru5: the deploy date in GraphQL is not being saved for the token but it is being saved in the database, so we will have to query from the DB.
I did push a fix to store the deploy date of the token when it is being created.

RubenSousaDinis: I'm saying here: https://github.com/talentprotocol/mvp/pull/506/files#diff-1d9573cea53b288a1cdca348e33e3fda347de76e516ec8e13cf2dcc7d1709abfR59.

app/packs/src/components/talent/TalentPage.jsx
```diff
@@ -34,11 +35,29 @@ const TalentPage = ({ talents }) => {
...
+      0
+    ) / msDividend;
+  const startDate = endDate - varianceDays * dayInSeconds;
+
   const { loading, data } = useQuery(GET_TALENT_PORTFOLIO, {
     variables: {
       ids: localTalents
         .map((talent) => talent.token.contractId)
         .filter((id) => id),
+      startDate,
```
RubenSousaDinis, line 59: `startDate > deployDate ? startDate : deployDate`

app/packs/src/utils/viewHelpers.js
RubenSousaDinis: Ok, got it. We need to compute the average.

app/packs/src/components/talent/TalentPage.jsx
taru5: Since the deploy date will not be same for all tokens, we cannot rely on this check here

taru5: Since the deploy date will not be same for all tokens, we cannot rely on this check here

RubenSousaDinis: I see. Got it!

RubenSousaDinis (approved):
Nice job1

PR #510, opened 2022-05-24
Add discovery row page

RubenSousaDinis:
## Summary

This PR adds the discovery row pages. We're adding them to showcase talent inside particular areas, such as web3, desing, development, etc. The pages should be public accessible.

## Notion link

https://talentprotocol.notion.site/Improve-Discover-Page-Automation-2645e05c601945d2b1cffa339d4c8d10

### DIFF
--- a/app/blueprints/discovery_row_blueprint.rb
+++ b/app/blueprints/discovery_row_blueprint.rb
@@ -1,7 +1,13 @@
 class DiscoveryRowBlueprint < Blueprinter::Base
-  fields :id, :badge, :badge_link, :slug, :title, :description, :logo_url, :talents_count
+  fields :id, :slug
 
   view :normal do
+    fields :badge, :badge_link, :title, :description, :logo_url, :talents_count
+  end
+
+  view :with_talents do
+    include_view :normal
+
     association :talents, blueprint: TalentBlueprint, view: :normal
   end
 end
--- a/app/blueprints/discovery_row_blueprint.rb
+++ b/app/blueprints/discovery_row_blueprint.rb
@@ -1,5 +1,5 @@
 class DiscoveryRowBlueprint < Blueprinter::Base
-  fields :id, :title, :badge, :badge_link
+  fields :id, :badge, :badge_link, :slug, :title
 
   view :normal do
     association :talents, blueprint: TalentBlueprint, view: :normal
--- a/app/blueprints/discovery_row_blueprint.rb
+++ b/app/blueprints/discovery_row_blueprint.rb
@@ -3,6 +3,8 @@ class DiscoveryRowBlueprint < Blueprinter::Base
 
   view :normal do
     fields :badge, :badge_link, :title, :description, :logo_url, :talents_count
+
+    association :tags, blueprint: TagBlueprint
   end
 
   view :with_talents do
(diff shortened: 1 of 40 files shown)


app/services/talents/search.rb
```diff
@@ -16,17 +18,31 @@ def call
...
+
+      users = users.where(
+        "users.username ilike :keyword " \
+        "OR users.display_name ilike :keyword " \
+        "OR tokens.ticker ilike :keyword " \
+        "OR tags.description ilike :keyword ",
+        keyword: "%#{keyword}%"
+      )
+
+      talents.where(user: users.distinct.select(:id))
```
verdefred, line 45: what's the reason for changing this from a talent query to a user and a talent one?

app/controllers/discovery_controller.rb
```diff
@@ -28,9 +29,36 @@ def index
...
+
+    respond_to do |format|
+      format.html
+      format.json {
+        render(
+          json: @talents,
+          status: :ok
+        )
+      }
+    end
```
verdefred, line 56: I think we don't need this, right?

app/services/talents/search.rb
RubenSousaDinis: The distinct does not work with the setseed. This way I do the distinct query in the users query and treat the talent query as it was before.

app/blueprints/discovery_row_blueprint.rb
```diff
@@ -1,5 +1,5 @@
 class DiscoveryRowBlueprint < Blueprinter::Base
-  fields :id, :title, :badge, :badge_link
+  fields :id, :badge, :badge_link, :slug, :title, :description, :logo_url, :talents_count
```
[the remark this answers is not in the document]
RubenSousaDinis: I'm a bit against writing code that might be needed in the future. I will include them in the normal view for now and when we need another view we can refactor.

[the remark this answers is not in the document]
RubenSousaDinis: Changed!

app/controllers/discovery_controller.rb
RubenSousaDinis: In the discovery rows we can also search and we use this endpoint to also return the json results.

verdefred: makes sense, missed that. thanks!

app/packs/src/components/discovery/discovery_rows.jsx
```diff
@@ -69,17 +73,17 @@ const DiscoveryRows = ({ discoveryRows, updateFollow }) => {
       return (
         <a
           href={row.badge_link}
-          className="cursor-pointer ml-2 "
+          className="cursor-pointer ml-2"
           target="_blank"
         >
-          <Tag className="bg-primary text-white cursor-pointer">
+          <Tag className=" cursor-pointer" mode={theme.mode()}>
```
verdefred, line 79: I don't think the tag component needs the mode anymore. it should work the same without it

RubenSousaDinis: removed!

verdefred (dismissed):
Great work 🚀

verdefred (approved):

PR #518, opened 2022-05-26
Add profile type to user

verdefred:
## Summary
We decided to add a profile_type enum field to the user so we can take care of the flow from Supporter to Talent. This will need to have some intermediary steps like applying, waiting for approval and getting an approval

## Notion link

https://www.notion.so/talentprotocol/Upgrade-to-Talent-Account-02e0b226fffa4d52829bb16b4ace9926

## How to test?

Nothing should really change in this PR. Everything should work the same, it's just the foundation for the next PRs. We need to run the rake task before going forward with the next PRs so we always have the correct data live

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -35,6 +35,14 @@ class User < ApplicationRecord
 
   VALID_ROLES = ["admin", "basic"].freeze
 
+  enum profile_type: {
+    supporter: "supporter",
+    applying: "applying",
+    waiting_for_approval: "waiting_for_approval",
+    approved: "approved",
+    talent: "talent"
+  }
+
   module Delivery
     TYPES = [
       MessageReceivedNotification,
--- a/app/services/supporter/upgrade_to_talent.rb
+++ b/app/services/supporter/upgrade_to_talent.rb
@@ -41,10 +41,6 @@ class Supporter::UpgradeToTalent
   end
 
   def update_profile_type(user, applying)
-    if applying
-      user.update(profile_type: "applying")
-    else
-      user.update(profile_type: "talent")
-    end
+    user.update(profile_type: applying ? "applying" : "talent")
   end
 end
--- a/app/services/supporter/upgrade_to_talent.rb
+++ b/app/services/supporter/upgrade_to_talent.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Supporter::UpgradeToTalent
-  def call(user_id:)
+  def call(user_id:, applying: false)
     user = User.find(user_id)
 
     return false if user.talent.present?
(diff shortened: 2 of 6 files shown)


db/migrate/20220524112720_add_profile_type_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddProfileTypeToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :profile_type, :string, default: "supporter"
```
RubenSousaDinis, line 3: add `null: false`?

app/services/supporter/upgrade_to_talent.rb
```diff
@@ -38,4 +39,12 @@ def copy_information_to_talent(user)
     user.talent.banner_data = user.investor.banner_data
     user.talent.save!
   end
+
+  def update_profile_type(user, applying)
+    if applying
+      user.update(profile_type: "applying")
+    else
+      user.update(profile_type: "talent")
```
RubenSousaDinis, line 47: One liner? 

```suggestion
      user.update(profile_type: applying ? "applying" : "talent")
```

db/migrate/20220524112720_add_profile_type_to_users.rb
verdefred: good catch, I'll make the change

app/services/supporter/upgrade_to_talent.rb
verdefred: Nice!

lib/tasks/tmp/update_users_profile_type.rake
```diff
@@ -0,0 +1,11 @@
+namespace :users do
+  task update_profile_type: :environment do
+    puts "updating users"
+    User.find_each do |user|
+      if user.profile_type == "investor" && user.talent.present?
+        user.update(profile_type: "talent")
```
RubenSousaDinis, line 6: I think we should fail if something goes wrong here. At least for us to be aware something failed.

```suggestion
        user.update!(profile_type: "talent")
```

app/models/user.rb
```diff
@@ -35,6 +35,14 @@ class User < ApplicationRecord
 
   VALID_ROLES = ["admin", "basic"].freeze
 
+  enum profile_type: {
+    supporter: "supporter",
+    applying: "applying",
+    waiting_for_approval: "waiting_for_approval",
+    approved: "approved",
```
RubenSousaDinis, line 42: what's the difference between "approved" and "talent"?

verdefred: Approved means someone has approved you so you can launch your token. Talent means you have been approved and have launched your token

RubenSousaDinis: ah ok! makes sense!

RubenSousaDinis (approved):
Well done! Nice job!

PR #523, opened 2022-05-27
Return 406 not acceptable when a json request is made on a html only endpoint

RubenSousaDinis:
## Summary

This PR prevents json requests to html only endpoints to raise internal server errors. From now on it will return: 406 Not Acceptable

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -18,7 +18,16 @@ class ApplicationController < ActionController::Base
     if Rails.env.development? || Rails.env.test?
       raise ActionController::RoutingError.new("We didn't find any routes that match your request.")
     else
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -15,10 +15,19 @@ def render_404
   end
 
   def route_not_found
-    if Rails.env.development? || Rails.env.test?
+    if Rails.env.development? && Rails.env.test?
```
verdefred, line 18: I don't understand this change. an environment is never both, right?

RubenSousaDinis: good catch. I did it for testing the change!

verdefred (approved):

PR #529, opened 2022-05-30
Add admin process to allow users to launch their token

verdefred:
## Summary
We add a new option to the search dropdown on the /talents page, only for admins, so they can search by talents waiting for approval to launch their token.
Then, inside the user's profile, there's an "approve" button (when before there was a "buy button") so the user can be approved. after that the user can launch their own token

## Notion link

https://www.notion.so/talentprotocol/Upgrade-to-Talent-Account-02e0b226fffa4d52829bb16b4ace9926

### DIFF
--- a/app/controllers/api/v1/talent_controller.rb
+++ b/app/controllers/api/v1/talent_controller.rb
@@ -30,7 +30,7 @@ class API::V1::TalentController < ApplicationController
   end
 
   def update
-    if talent.id != current_user.talent.id
+    if !current_user.admin? && talent.id != current_user.talent&.id
       return render json: {error: "You don't have access to perform that action"}, status: :unauthorized
     end
 
--- a/app/services/talents/search.rb
+++ b/app/services/talents/search.rb
@@ -1,13 +1,14 @@
 module Talents
   class Search
-    def initialize(filter_params: {}, sort_params: {}, discovery_row: nil)
+    def initialize(filter_params: {}, sort_params: {}, discovery_row: nil, admin: false)
       @filter_params = filter_params
       @sort_params = sort_params
       @discovery_row = discovery_row
+      @admin = admin
     end
 
     def call
-      talents = Talent.base.joins(:user, :token)
+      talents = admin ? Talent.joins(:user, :token) : Talent.base.joins(:user, :token)
 
       talents = filter_by_discovery_row(talents) if discovery_row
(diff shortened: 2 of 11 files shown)


app/controllers/api/v1/talent_controller.rb
```diff
@@ -30,7 +30,7 @@ def show
   end
 
   def update
-    if talent.id != current_user.talent.id
+    if !current_user.admin? && talent.id != current_user.talent&.id
```
RubenSousaDinis, line 33: Should it be or here?

```suggestion
    if !current_user.admin? || talent.id != current_user.talent&.id
```

RubenSousaDinis, line 33: Should it be or here?
```suggestion
    if !current_user.admin? || talent.id != current_user.talent&.id
```

app/services/talents/search.rb
```diff
@@ -57,6 +57,8 @@ def filter_by_status(talents)
           .active
           .where("tokens.deployed_at > ?", 1.month.ago)
           .order("tokens.deployed_at ASC")
+      elsif filter_params[:status] == "Pending approval"
+        Talent.joins(:user).where(user: {profile_type: "waiting_for_approval"})
```
RubenSousaDinis, line 61: we lose the previous filtering by doing this.

```suggestion
        talents.joins(:user).where(user: {profile_type: "waiting_for_approval"})
```

RubenSousaDinis (commented):
Left some comments

app/controllers/api/v1/talent_controller.rb
verdefred: I don't think so. even if a user is not an admin they should be able to change their own profile. this was added in the case an admin wants to change another user, which is now possible (to change the profile_type)

app/services/talents/search.rb
verdefred: Good catch. unfortunately your suggestion doesn't work because we do Talent.base before, which filters out the private talents. in this case we need both. I'll dm you to discuss the best solution

app/controllers/api/v1/talent_controller.rb
RubenSousaDinis: Yup, got it!

RubenSousaDinis (approved):
Looks good! Nice job!

PR #530, opened 2022-05-31
Fix webpack hot reloading

verdefred:
## Summary

When the package webpacker-dev-server was updated, because of a CVE found, we lost the hot reloading functionality, because the new version runs the server on a different port.
To fix this we need to change the port for the new one used


RubenSousaDinis (approved):
Awesome!

PR #532, opened 2022-05-31
Partnership pages

RubenSousaDinis:
## Summary

This PR adds partnerships to discovery row pages. We will now display more information (logo and links) when the discovery row is associated with a partnership.

It also automates the generation of discovery rows based on invites. When the invite is associated with a discovery row when a new user creates the account the discovery row will be created by default.

## Notion link

https://www.notion.so/talentprotocol/Improve-Discover-Page-Automation-2645e05c601945d2b1cffa339d4c8d10


app/services/invites/create.rb
```diff
@@ -0,0 +1,53 @@
+module Invites
+  class Create
```
RubenSousaDinis, line 2: Refactored. Previous file: https://github.com/talentprotocol/mvp/pull/532/files#diff-1a7c37d5abb9c90c937c2de4155dfc65f91c0935dfb77b45131395ce83e3a366

app/services/users/create.rb
```diff
@@ -0,0 +1,145 @@
+module Users
+  class Create
```
RubenSousaDinis, line 2: Refactored. Previous file: https://github.com/talentprotocol/mvp/pull/532/files#diff-39b9382a72b38dcd3b05ba527fe283e7ec24aff254555c1ddab86e8752e4c3ae

app/services/invites/create_partnership.rb
```diff
@@ -0,0 +1,43 @@
+module Invites
+  class CreatePartnership
```
RubenSousaDinis, line 2: Not used in the code. Created only to ease the creation of Partnership invites.

verdefred (dismissed):
🚀

verdefred (approved):

PR #548, opened 2022-06-07
Add hide profile to talent model

francisco-leal:
## Summary

We need a way to force hiding a profile before disabling it


db/migrate/20220607144915_add_hide_profile_to_talent.rb
```diff
@@ -0,0 +1,5 @@
+class AddHideProfileToTalent < ActiveRecord::Migration[6.1]
+  def change
+    add_column :talent, :hide_profile, :boolean, default: false
```
verdefred, line 3: can we add `null: false` since the default is false (I'm guessing we don't need the field to be null)

PR #549, opened 2022-06-07
30 day market cap improvements

RubenSousaDinis:
## Summary

This PR removes the text highlight when the variance is 0. It also removes it when it is lower than 0.009.

### DIFF
--- a/app/packs/src/components/portfolio/components/Supporting.jsx
+++ b/app/packs/src/components/portfolio/components/Supporting.jsx
@@ -19,6 +19,8 @@ import Table from "src/components/design_system/table";
 import Link from "src/components/design_system/link";
 import { OrderBy } from "src/components/icons";
 
+import { parsedVariance } from "src/utils/viewHelpers";
+
 const concatenateSupportingAddresses = (supporting) =>
   `?tokens[]=${supporting.map((s) => s.contract_id).join("&tokens[]=")}`;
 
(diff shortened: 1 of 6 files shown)


app/packs/src/components/portfolio/components/Supporting.jsx
```diff
@@ -251,6 +253,12 @@ const Supporting = ({
     }
   };
 
+  const varianceClassNames = (variance) => {
+    if (!variance || Math.abs(variance) < 0.009) return "";
```
verdefred, line 257: shouldn't it be `< 0.01`?

app/packs/src/components/talent/TalentTableListMode.jsx
```diff
@@ -146,6 +148,17 @@ const TalentTableListMode = ({
     setShowDropdown(false);
   };
 
+  const varianceClassNames = (talent) => {
+    if (
+      !talent.token.contractId ||
+      !talent.marketCapVariance ||
+      Math.abs(talent.marketCapVariance) < 0.009
```
verdefred, line 155: same here

app/packs/src/components/portfolio/components/Supporting.jsx
RubenSousaDinis: good point, fixed!

verdefred (approved):

PR #561, opened 2022-06-09
Create user profile type changes

verdefred:
## Summary
We want to have more control over when the users change their profile type or when someone on the admin team approves someone. For this we're creating a new model that tracks that

### DIFF
--- a/app/models/user_profile_type_change.rb
+++ b/app/models/user_profile_type_change.rb
@@ -1,4 +1,4 @@
 class UserProfileTypeChange < ApplicationRecord
-  belongs_to :user, class_name: "User"
-  belongs_to :who_dunnit, class_name: "User", optional: true
+  belongs_to :user, class_name: "User", foreign_key: "user_id"
+  belongs_to :who_dunnit, class_name: "User", foreign_key: "who_dunnit_id"
 end
--- a/app/models/user_profile_type_change.rb
+++ b/app/models/user_profile_type_change.rb
@@ -1,0 +1,4 @@
+class UserProfileTypeChange < ApplicationRecord
+  belongs_to :user, class_name: "User"
+  belongs_to :who_dunnit, class_name: "User", optional: true
+end
--- a/app/services/api/update_talent.rb
+++ b/app/services/api/update_talent.rb
@@ -1,9 +1,9 @@
 class API::UpdateTalent
-  attr_reader :talent, :current_user, :success
+  attr_reader :talent, :user, :success
 
-  def initialize(talent, current_user)
+  def initialize(talent, user)
     @talent = talent
-    @current_user = current_user
+    @user = user
     @success = false
   end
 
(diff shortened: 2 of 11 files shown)


app/models/user_profile_type_change.rb
```diff
@@ -0,0 +1,4 @@
+class UserProfileTypeChange < ApplicationRecord
+  belongs_to :user, class_name: "User"
+  belongs_to :who_dunnit, class_name: "User", optional: true
```
RubenSousaDinis, line 3: Would it ever happen that a change is done and the who_dunnit is null?

app/services/api/update_talent.rb
```diff
@@ -1,8 +1,9 @@
 class API::UpdateTalent
-  attr_reader :talent, :success
+  attr_reader :talent, :current_user, :success
```
RubenSousaDinis, line 2: Personal preference, but in the business area of the application (services) we shouldn't have the current_user notion. It should only live in the requests area. Here it should simply be a user.

db/migrate/20220609162621_create_user_profile_type_changes.rb
```diff
@@ -0,0 +1,15 @@
+class CreateUserProfileTypeChanges < ActiveRecord::Migration[6.1]
+  def change
+    create_table :user_profile_type_changes do |t|
+      t.integer :user_id
```
RubenSousaDinis, line 4: we're not creating  a foreign_key with this approach. Also it should be not nullable.

spec/services/users/update_profile_type.rb
```diff
@@ -0,0 +1,52 @@
+require "rails_helper"
+
+RSpec.describe Users::UpdateProfileType do
+  include ActiveJob::TestHelper
```
RubenSousaDinis, line 4: not needed

spec/services/users/update_profile_type.rb
```diff
@@ -0,0 +1,52 @@
...
+      expect(result.user_id).to be(user.id)
+      expect(result.who_dunnit_id).to be(user.id)
+      expect(result.new_profile_type).to eq(new_profile_type)
+    end
+  end
+
+  context "when an invalid profile_type is provided" do
+    let(:new_profile_type) { "gibberish" }
+
+    it "it updates the profile_type" do
```
RubenSousaDinis, line 36:
```suggestion
    it "it fails to update the profile_type" do
```

RubenSousaDinis (commented):
Some comments to be considered.

app/services/api/update_talent.rb
verdefred: makes sense, I'll make the change

RubenSousaDinis (approved):

RubenSousaDinis (approved):

PR #562, opened 2022-06-09
Talent impersonation

MinervaMcG:
## Summary

This PR enables admins to impersonate any user in the system. It allows admins to view all pages a user would see and also allow them to edit profiles to figure out what users are missing in their profile and help them get through.
We keep track of all impersonations and added [Paper Trail](https://github.com/paper-trail-gem/paper_trail) in this PR to keep audits of the changes in the profile.

## Notion link

https://talentprotocol.notion.site/Allow-admins-to-Impersonate-users-7492fbf4f176491381f65f8a945506ae)

## How to test?

The URL to start impersonating a user is `/impersonate/{username}`. Impersonation is only enabled if the user is an admin and lasts for 30 minutes.


test/models/impersonation_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+class ImpersonationTest < ActiveSupport::TestCase
```
RubenSousaDinis, line 3: Remove file

RubenSousaDinis (commented):
Some files are not needed and would be good to add some specs

app/helpers/impersonations_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ImpersonationsHelper
```
RubenSousaDinis, line 1: Can be removed

app/models/user.rb
```diff
@@ -33,6 +35,10 @@ class User < ApplicationRecord
   # Rewards
   has_many :rewards
 
+  # Impersonation
+  has_many :impersonators, foreign_key: :impersonator_id, class_name: "Impersonation"
+  has_many :impersonations, foreign_key: :impersonated_id, class_name: "Impersonation"
+
```
RubenSousaDinis, line 41: I'm not seeing this being used anywhere. Why do we need it?

app/models/user.rb
```diff
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
+  has_paper_trail ignore: [:updated_at]
```
RubenSousaDinis, line 2: Why are we ignoring the updated_at field?

db/migrate/20220526171942_create_impersonations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateImpersonations < ActiveRecord::Migration[6.1]
+  def change
+    create_table :impersonations do |t|
+      t.integer :impersonator_id, null: false
+      t.integer :impersonated_id, null: false
```
RubenSousaDinis, line 5: Since this are foreign keys we should treat them accordingly. Missing index and foreign key

app/controllers/messages_controller.rb
```diff
@@ -59,13 +59,13 @@ def create
   end
 
   def send_to_all_supporters
-    if message_params[:message].blank?
+    if message_params[:message].blank? || is_user_impersonated?
```
RubenSousaDinis, line 62: This prevention is good! Just wondering if we should have a shared validation method in application controller that halts the execution of some endpoints if the user is being impersonated. In the controller we can use a before action to run the method. WDYT?

app/controllers/messages_controller.rb
```diff
@@ -3,27 +3,27 @@ class MessagesController < ApplicationController
   before_action :set_user, only: [:index]
 
   def index
```
RubenSousaDinis, line 5: Wondering if we should return empty results on this endpoint. We will show private messages if we don't do it. We can disable the messages tab, but the BE protection is always better.

RubenSousaDinis (changes requested):
Some changes required. Overall is looking good. Also tested locally and I have some comments:

- [ ] We should have a way to stop impersonation the user (a button in the header will do it)
- [ ] We should prevent private information (messages and probably notifications) to be displayed when impersonating the user

app/models/user.rb
MinervaMcG: Since we update the user on every activity or page visit, the updated_at column is updated, without actually updating another field. In order to restrict the audits in the case, I ignored the updated_at column.

app/models/user.rb
MinervaMcG: We can ignore these at this point

app/controllers/impersonations_controller.rb
```diff
@@ -0,0 +1,32 @@
+class ImpersonationsController < ApplicationController
+  def show
```
RubenSousaDinis, line 2: this should be a post request since we're changing the server state by creating records

app/controllers/application_controller.rb
```diff
@@ -51,6 +73,10 @@ def talent_id_param
   end
 
   def current_user_watchlist
-    current_user ? current_user.following.pluck(:user_id) : []
+    current_acting_user ? current_acting_user.following.pluck(:user_id) : []
+  end
+
+  def user_from_impersonated_cookie
+    User.where(username: cookies.signed[:impersonated]).first
```
RubenSousaDinis, line 80:
```suggestion
    User.find_by(username: cookies.signed[:impersonated])
```

app/controllers/impersonations_controller.rb
```diff
@@ -0,0 +1,32 @@
+class ImpersonationsController < ApplicationController
```
RubenSousaDinis, line 1: Can you add request specs for the endpoints below?

RubenSousaDinis (commented):
I stopped seeing the button to impersonate. Left some more comments.

app/controllers/impersonations_controller.rb
MinervaMcG: If we change it to post, the admin won't be able to access this URL directly from the browser to start impersonating a user.

app/assets/stylesheets/impersonations.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the impersonations controller here.
```
RubenSousaDinis, line 1: Remove empty file

RubenSousaDinis (commented):
Some small comments. Overall looks good. I also tested and it's working properly! Well done!

spec/requests/api/v1/impersonations_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  describe "#destroy" do
+    subject(:delete_impersonation_request) { delete api_v1_impersonation_path(id: impersonated_user.id, as: current_user) }
+
+    context "when user is admin" do
+      let(:current_user) { create :user, role: "admin" }
+
+      it "should destroy cookie when user is impersonated" do
+        create_impersonation_request
```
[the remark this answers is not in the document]
MinervaMcG: It is not allowing us to add a signed cookie directly in the scenario, so had to call the create request first

db/migrate/20220624180204_add_indexes_to_impersonations.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexesToImpersonations < ActiveRecord::Migration[6.1]
```
MinervaMcG, line 1: @RubenSousaDinis Added foreign keys in this migration

spec/requests/api/v1/impersonations_spec.rb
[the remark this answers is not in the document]
RubenSousaDinis: Ok, I was also not able to mock signed cookies

RubenSousaDinis (approved):
Great work!

PR #609, opened 2022-07-13
Fix: Load supporter count from db

francisco-leal:
## Summary

After the smart contract migration the supporterCount field on thegraph is no longer up to date. We need to use the value from our database that we have from our TalentSupporter sync job.


RubenSousaDinis (approved):
Well done!

PR #616, opened 2022-07-13
Small fixes

francisco-leal:
## Summary

- Makes all profiles start as talent with token & perks locked
- Removes private profiles, the profile becomes public after finishing the quest "Talent Profile" or you're approved to launch the token
- A few copy tweaks

## Notion link

https://www.notion.so/talentprotocol/Bug-Report-Quick-Fixes-b51a97beb8a341f6a2c8a962671087b0


verdefred (approved):
🚀

PR #623, opened 2022-07-19
Add auto complete to tags

verdefred:
## Summary
Show total of users with each tag and add auto complete to tags dropdown on the Edit Profile

## Notion link

https://www.notion.so/talentprotocol/Tags-Show-totals-and-add-auto-complete-b7fe33e283a441369574cbe1fee87bd3

### DIFF
--- a/app/controllers/api/v1/tags_controller.rb
+++ b/app/controllers/api/v1/tags_controller.rb
@@ -3,7 +3,7 @@ class API::V1::TagsController < ApplicationController
     tags =
       Tag
         .visible
-        .joins(:user_tags)
+        .left_joins(:user_tags)
         .order("user_tags.count DESC")
         .where("description ilike ?", "%#{params[:description]}%")
         .group(:id)
--- a/app/controllers/api/v1/tags_controller.rb
+++ b/app/controllers/api/v1/tags_controller.rb
@@ -1,0 +1,12 @@
+class API::V1::TagsController < ApplicationController
+  def index
+    tags =
+      Tag
+        .order(:id)
+        .where(hidden: false)
+        .where("description ilike ?", "%#{params[:description]}%")
+        .limit(7)
+
+    render json: TagBlueprint.render(tags), status: :ok
+  end
+end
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/tags_controller.rb
```diff
@@ -0,0 +1,12 @@
+class API::V1::TagsController < ApplicationController
+  def index
+    tags =
+      Tag
+        .order(:id)
+        .where(hidden: false)
```
RubenSousaDinis, line 6: we have a scope: `visible`

RubenSousaDinis (commented):
Left some comments. Great work!

app/controllers/api/v1/tags_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+        .where("description ilike ?", "%#{params[:description]}%")
+        .limit(7)
```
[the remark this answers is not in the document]
verdefred: it was something discussed with design. they feel that more that 7 will be too much for a dropdown but I can increase it a bit, I guess

app/controllers/api/v1/tags_controller.rb
```diff
@@ -0,0 +1,14 @@
+class API::V1::TagsController < ApplicationController
+  def index
+    tags =
+      Tag
+        .visible
+        .joins(:user_tags)
```
RubenSousaDinis, line 6: should we use left join here? If a tag has no usage and it exists in the system it will not appear.

RubenSousaDinis (commented):
Just one small comment. We are ready to go after this one!

RubenSousaDinis (dismissed):
Well done!

app/packs/src/components/supporters/edit/About.jsx
```diff
@@ -178,11 +185,27 @@ const About = ({
     changeSharedState((prev) => ({ ...prev, [attribute]: value }));
   };
 
-  const changeTags = (tags) => {
+  const getTags = (tagName) => {
```
francisco-leal, line 188: let's add a short debounce here so we don't flood the server.

RubenSousaDinis (approved):
Well done!

PR #627, opened 2022-07-20
Fix update profile type

verdefred:
## Summary
Fix to the update profile type so we always have the most updated user in memory.
Doesn't make sense to go back once a user is approved so now it's not possible


RubenSousaDinis (approved):
Well done!

PR #646, opened 2022-08-05
Adds ens and wallet routes

sujink1999:
## Summary

This PR adds user profile routes for wallet_id and ens

## Notion link

 https://talentprotocol.notion.site/Wallet-addresses-pages-cfdb492a3210490aa2ca381fd8e64372

## How to test?

Migrate the DB changes
Try visiting a profile with the wallet id
Try updating a ens domain for a user from the Edit profile section and navigate to the profile using ens

### DIFF
--- a/app/packs/src/components/talent/Edit/About.jsx
+++ b/app/packs/src/components/talent/Edit/About.jsx
@@ -186,7 +186,9 @@ const About = (props) => {
     if (props.user.wallet_id) {
       setFetchENS((prev) => ({ ...prev, loading: true}));
       const name = await getENSFromAddress(props.user.wallet_id);
-      name && changeUserAttribute("ens", name);
+      if(name){
+        changeUserAttribute("ens_domain", name);
+      }
     }
     setFetchENS((prev) => ({ loading: false, hasAlreadyFetched: true}));
   };
@@ -399,7 +401,7 @@ const About = (props) => {
           onChange={(e) =>
             changeTalentProfileAttribute("occupation", e.target.value)
           }
-          value={props.user.ens || (fetchENS["hasAlreadyFetched"] ? "No domain available" : "")}
+          value={props.user.ens_domain || (fetchENS["hasAlreadyFetched"] ? "No domain available" : "")}
           disabled
           className="flex-fill mr-3"
         />
--- a/app/packs/src/components/talent/Edit/About.jsx
+++ b/app/packs/src/components/talent/Edit/About.jsx
@@ -22,6 +22,7 @@ import Checkbox from "src/components/design_system/checkbox";
 import { ArrowRight } from "src/components/icons";
 import LoadingButton from "src/components/button/LoadingButton";
 import Divider from "src/components/design_system/other/Divider";
+import { getENSFromAddress } from "../../../onchain/utils";
 
 import cx from "classnames";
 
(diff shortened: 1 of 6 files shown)


app/packs/src/onchain/utils.js
```diff
@@ -7,3 +8,14 @@ export const parseAndCommify = (value, nrOfDecimals = 2) => {
 export const ipfsToURL = (ipfsAddress) => {
   return "https://ipfs.io/" + ipfsAddress.replace("://", "/");
 };
+
+export const getENSFromAddress = async (walletAddress) => {
+  let name = null;
+  try {
+    const provider = new ethers.providers.EtherscanProvider();
+    const name = await provider.lookupAddress(walletAddress);
```
RubenSousaDinis, line 16: Constant already defined above

```suggestion
    name = await provider.lookupAddress(walletAddress);
```

app/packs/src/onchain/utils.js
```diff
@@ -1,4 +1,5 @@
 import { ethers } from "ethers";
+import Addresses from "./addresses.json";
```
RubenSousaDinis, line 2: Not used

db/migrate/20220805060322_add_ens_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddEnsToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :ens, :string
```
RubenSousaDinis, line 3: I would call this field `ens_domain`. I think it will be more explicit for developers starting in web3

RubenSousaDinis (commented):
Left some comments. Overall the code looks good!

app/packs/src/components/talent/Edit/About.jsx
```diff
@@ -176,6 +182,16 @@ const About = (props) => {
     }
   };
 
+  const updateUserENS = async () => {
+    if (props.user.wallet_id) {
+      setFetchENS((prev) => ({ ...prev, loading: true}));
+      const name = await getENSFromAddress(props.user.wallet_id);
+      name && changeUserAttribute("ens", name);
```
francisco-leal, line 189: I would consider this personal preference but I'm not a huge fan of doing this sort of things in-line. Javascript can already be a bit messy ahah so I always favour readability. What do you think of splitting it into an if
```js
if (name) {
  changeUserAttribute("ens", name);
}
```

db/migrate/20220805060322_add_ens_to_users.rb
francisco-leal: I agree with going with `ens_domain`. I could see a future where we even turn this into its own model so we can support multiple domains, like nomspace

francisco-leal (commented):
nice work!

RubenSousaDinis (changes requested):
I was testing this and found an issue. You need to add a constraint to the route (https://github.com/talentprotocol/web-dapp/blob/c083b6dcede26dfc3a78531e3d557f8a3c3bf28a/config/routes.rb#L120):

 `get "/u/:username" => "users#show", :as => "user", :constraints => {username: /[^\/]+/}`

RubenSousaDinis:
> I was testing this and found an issue. You need to add a constraint to the route (
> 
> https://github.com/talentprotocol/web-dapp/blob/c083b6dcede26dfc3a78531e3d557f8a3c3bf28a/config/routes.rb#L120
> 
> ):
> `get "/u/:username" => "users#show", :as => "user", :constraints => {username: /[^\/]+/}`

Basically the `.` is a special character in the urls formats and by default rails consider it a format splitting the param into two different params.

RubenSousaDinis (approved):
Great work!

PR #648, opened 2022-08-08
feat: LinkedIn OAuth login

shirleyodeng:
## Summary

This PR;
- adds a "Sign In with LinkedIn" button on the home / log in page.
- adds a LinkedIn client class
- adds a LinkedIn OAuth callback endpoint
- adds a LinkedIn OAuth handler that either finds or creates a user via the email address

Screenshot 2022-08-08 at 09 07 56

## Notion link

https://talentprotocol.notion.site/Sign-in-with-LinkedIn-API-Integration-1c7b740cbb4d4626a7b9812d6e592e61

## How to test?

- You'll need the LinkedIn credentials (Client ID & Secret) via the [LinkedIn Developer Portal](https://www.linkedin.com/developers/apps/204863218/settings)

## Notes

- Was not able to retrieve the "headline" field from LinkedIn as this is part of the [Basic Profile scope](https://docs.microsoft.com/en-us/linkedin/shared/references/v2/profile/basic-profile?view=li-lms-2022-07) that we do not have permissions for.

Screenshot 2022-08-08 at 09 09 41


app/uploaders/image_uploader.rb
```diff
@@ -7,7 +7,7 @@ class ImageUploader < Shrine
     validate_max_size 10 * 1024 * 1024
 
     # Validate that the image format is an expected one
-    validate_extension %w[jpg jpeg png]
+    validate_extension %w[jpg jpeg png] unless context[:omniauth]
```
shirleyodeng, line 10: LinkedIn does not provide an extension; https://stackoverflow.com/questions/68023473/shrine-validation-of-file-extension-does-not-allow-facebook-image

app/packs/src/components/button/LinkedinSignInButton.jsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import LinkedinSignIn from "images/linkedin-sign-in.png";
+
+const LinkedinSignInButton = ({
+  clientId,
+  redirectUri,
+}) => {
+  const baseAuthorizationUri = "https://www.linkedin.com/oauth/v2/authorization"
+  const scope = "r_liteprofile%20r_emailaddress"
```
RubenSousaDinis, line 9: It might be better for us to request access to the full profile now. It might require extra work when we request more info (timeline, skills, etc.) in the future. I remember the access token is valid for 2 months and we might ended up needing it the full access permission. @francisco-leal WDYT?

app/services/linkedin/oauth_handler.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  def lite_profile
+    @lite_profile ||= JSON.parse(lite_profile_request.body)
+  end
+
+  def upload_profile_picture(user)
+    talent = user.talent
+    return unless talent
+    return if talent.profile_picture_url
```
RubenSousaDinis, line 102: We need to do the same for investors here. They might also sign in using linkedin. A user can be a talent (picture coming from the talent model) or investor (picture coming from the investor model). Check here how we retrieve the user image: https://github.com/talentprotocol/web-dapp/blob/c083b6dcede26dfc3a78531e3d557f8a3c3bf28a/app/models/user.rb#L178-L180

app/services/users/create.rb
```diff
@@ -6,12 +6,18 @@ def initialize
...
     end
 
-    def call(email:, username:, password:, invite_code:, theme_preference:)
+    def call(
+      email:,
+      password:,
+      username:,
+      invite_code: nil,
+      linkedin_id: nil,
+      theme_preference: "light"
```
RubenSousaDinis, line 15: Great!

RubenSousaDinis (commented):
Great piece of work! 🚀 I left some comments for us to discuss, but overall the code looks great! Thanks for the specs!

app/services/linkedin/oauth_handler.rb
shirleyodeng: Ahhh missed the investor model! I'll add the picture profile to the investor as well 👍

app/services/linkedin/oauth_handler.rb
```diff
@@ -0,0 +1,128 @@
...
+    raise BadResponseError, result[:error] unless result[:success]
+
+    result[:user].tap do |user|
+      user.confirm_email
+      AddUsersToMailerliteJob.perform_later(user.id)
+    end
+  end
+
+  def username
+    name = lite_profile.slice("localizedFirstName", "localizedLastName").values.join("").downcase
```
RubenSousaDinis, line 89: We also need to remove invalid chars from the username. For instance my name is not valid: `Rúben Dinis` due to the accentuation

app/services/linkedin/oauth_handler.rb
```diff
@@ -0,0 +1,128 @@
...
+  def lite_profile_request
+    @lite_profile_request ||= client.retrieve_lite_profile(access_token)
+  end
+
+  def create_user
+    result = Users::Create.new.call(
+      email: email_address,
+      linkedin_id: lite_profile["id"],
+      password: nil,
+      username: username
```
RubenSousaDinis, line 77: We can also store the display name of the user with Linkedin first and last name

app/services/linkedin/oauth_handler.rb
shirleyodeng: Nice find, fixed!

app/services/linkedin/oauth_handler.rb
shirleyodeng: Added!

RubenSousaDinis (approved):

PR #678, opened 2022-08-30
Fetch CELO nfts from tatum api client

RubenSousaDinis:
## Summary

This PR adds the tatum api client to fetch celo nfts.

## Notion link

https://www.notion.so/talentprotocol/New-Profile-implementation-aee6a9f1073e4a0480b68d035c056177


db/schema.rb
```diff
@@ -13,7 +13,6 @@
 ActiveRecord::Schema.define(version: 2022_08_23_152927) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
verdefred, line 16: do we want to delete this?

lib/gnosis_chain_explorer/client.rb
```diff
@@ -1,23 +0,0 @@
...
-    private
-
-    def headers
-      {
-        "Content-Type": "application/json",
-        Accept: "application/json"
-      }
-    end
-  end
-end
```
verdefred, line 23: do we want to delete this?

lib/web3/api_proxy.rb
```diff
@@ -138,6 +138,29 @@ def moralis_nfts
...
+      tokens = response_body.map do |item|
+        item["metadata"].map do |token_metadata|
+          {
+            address: item["contractAddress"],
+            token_id: token_metadata["tokenId"],
+            amount: 1,
+            name: token_metadata["metadata"].present? ? token_metadata["metadata"]["name"] : nil,
+            symbol: token_metadata["metadata"].present? ? token_metadata["metadata"]["name"] : nil,
+            token_uri: token_metadata["url"].start_with?("http") ? token_metadata["url"] : nil,
+            metadata: token_metadata["metadata"].present? ? token_metadata["metadata"] : nil
```
verdefred, line 156: is there a reason to get all these things? I think we only need the address and the token id to then get the nft image, right?

verdefred (commented):
great work, just left a couple of questions

lib/gnosis_chain_explorer/client.rb
RubenSousaDinis: this is now in the web3 folder

lib/web3/api_proxy.rb
RubenSousaDinis: I think it doesn't hurt to get as much data as we can. Might be good for debugging

db/schema.rb
RubenSousaDinis: added back

lib/gnosis_chain_explorer/client.rb
verdefred: but was it duplicated in 2 places? I don't see it being added anywhere in this PR

RubenSousaDinis: Yes, it was probably a rebase issue. It was already in the web3 folder

verdefred (approved):
🚀

PR #682, opened 2022-09-02
Cache web3 data

RubenSousaDinis:
## Summary

This PR adds a service that will refresh tokens and nfts and store them in our database for caching purposes.

## Notion link

https://www.notion.so/talentprotocol/New-Profile-implementation-aee6a9f1073e4a0480b68d035c056177


db/migrate/20220830134506_create_erc721_tokens.rb
```diff
@@ -0,0 +1,23 @@
+class CreateErc721Tokens < ActiveRecord::Migration[6.1]
+  def change
+    create_table :erc721_tokens do |t|
+      t.string :address, null: false
+      t.string :name
+      t.string :symbol
+      t.string :url
+      t.json :metadata
+      t.string :token_id
+      t.string :amount
```
verdefred, line 10: since the erc721 is non fungible, does it make sense to have an amount? is this token the collection of NFTs or the NFT itself?

RubenSousaDinis: I've added the field since Moralis is returning it in their API. I've never seen any amount != 1 so far doing tests.

lib/web3/api_proxy.rb
```diff
@@ -1,5 +1,7 @@
-require "web3/moralis/client"
 require "web3/celo_explorer/client"
+require "web3/gnosis_chain_explorer/client"
```
verdefred, line 2: in the method gnosis_chain_explorer_client it should be Web3::GnosisChainExplorer::Client.new

verdefred (commented):

lib/web3/api_proxy.rb
RubenSousaDinis: It's ok, because we're already inside the Web3 module. Outside this module yes, we will need to add the `Web3::`.

verdefred (approved):

PR #685, opened 2022-09-04
feat: goal reminders

shirleyodeng:
## Summary

This PR adds two email reminders;
- for when a goal's due date is today
- for when a goal's due date is half-way

## Notion link

https://talentprotocol.notion.site/Reminders-for-Career-Goals-bb29ff212ddd4685a2d659ee553d4cde


app/views/user_mailer/send_goal_deadline_reminder_email.erb
```diff
@@ -0,0 +1,23 @@
...
+</h3>
+<p class="w-full primary-3 paragraph">
+  Hopefully "<%= @goal.title %>" goal is now done! Probably this is a good time to:
+</p>
+<ul class="mb-2 paragraph primary-3">
+  <li class="mb-2">
+    <%= link_to "Update your profile and roadmap", edit_profile_url(username: @user.username) %> 📝
+  </li>
+  <li>
+    Send an <%= link_to "accountability update to your token holders", "#" %> on your progress, and request their support.
```
shirleyodeng, line 12: @RubenSousaDinis; on Notion it says, `link to message all token holders` - is this the `messages_path`?

app/views/user_mailer/send_goal_deadline_reminder_email.erb
```diff
@@ -0,0 +1,23 @@
...
+    This is precisely why we built Talent Protocol in the first place 🤲
+  </li>
+</ul>
+<p class="mb-6 w-full primary-3 paragraph">
+  (for unverified folks only) Last but not least, consider <%= link_to "booking an onboarding", "#" %> call to get the verified checkmark on your profile ✅
```
shirleyodeng, line 17: @RubenSousaDinis; on Notion it says, `link to typeform` - can you provide the path/URL?

config/sidekiq.yml
```diff
@@ -9,6 +9,12 @@
 :limits:
   low: 1
 :schedule:
+  send_goal_deadline_reminder:
+    cron: "0 12 * * *" # Runs once every day at 12:00
+    class: EmailReminders::SendGoalDeadlineReminderJob
+  send_goal_halfway_reminder:
+    cron: "0 12 * * *" # Runs once every day at 12:00
+    class: EmailReminders::SendGoalHalfwayReminderJob
```
shirleyodeng, line 17: @RubenSousaDinis, users can only set the due date by the month (without the day). Is that intended or should I change this so the user can set the date?

Screenshot 2022-09-04 at 16 52 32

app/views/user_mailer/send_goal_deadline_reminder_email.erb
RubenSousaDinis: Link: https://talentprotocol.typeform.com/onboarding-call

app/views/user_mailer/send_goal_deadline_reminder_email.erb
RubenSousaDinis: We open the message all supporters on a modal. So I will say that for now it's ok to redirect to the messages path yes.

config/sidekiq.yml
francisco-leal: Setting a day is too specific for something like a goal, which is why we just ask for the month. For this particular case, we should use the last day of the month for any time calculations!

RubenSousaDinis: Should we run a rake task to update all goals to the last day of the month and start saving them that way from now on?

shirleyodeng: Sure, I'll add a rake task to this PR!

shirleyodeng: @RubenSousaDinis added > https://github.com/talentprotocol/web-dapp/pull/685/commits/a35f8cae9503f43a0dad6743404ab265cc6acfaa

config/sidekiq.yml
```diff
@@ -9,6 +9,12 @@
 :limits:
   low: 1
 :schedule:
+  send_goal_deadline_reminder:
+    cron: "0 12 L * *" # Runs on the last day of the month at 12:00
+    class: EmailReminders::SendGoalDeadlineReminderJob
```
shirleyodeng, line 14: @RubenSousaDinis I've changed this to only run on the last day of the month as that's when all the deadlines are

Screenshot 2022-09-08 at 08 38 10

RubenSousaDinis: Great!

RubenSousaDinis (approved):
Great work!

PR #686, opened 2022-09-05
Add pagination to invited users on earn page

verdefred:
## Summary
There are users with a lot of invited users on the earn page, which breaks the page. This adds pagination besides adding some includes that help while doing some queries


francisco-leal (approved):

RubenSousaDinis (approved):
Well done!

PR #714, opened 2022-10-02
Talent network filter

MinervaMcG:
## Summary

This PR adds two new filters to the Talent Page. 
1. By Celo Network
2. By Polygon Network
This enables searching for the talents that launched their token on the respective networks.

This PR also addresses an issue during the talent search which is when the user selects any filter option from the dropdown, the market cap, and progress of the user would disappear.


lib/web3/api_proxy.rb
```diff
@@ -17,6 +17,11 @@ class ApiClientRequestError < Error; end
     CELO_CHAIN = %w[celo 0xa4ec 42220]
     ETH_CHAIN = %w[eth 0x1 1]
     POLYGON_CHAIN = %w[polygon 0x89 137]
+
+    STAGING_CELO_CHAIN = %w[celo 0xaef3 44787]
```
RubenSousaDinis, line 21: I would call these constants `TESTNET_...` instead of `STAGING_...`. We might use the constants in more environments like dev

app/services/talents/search.rb
```diff
@@ -62,6 +66,12 @@ def filter_by_status(talents)
         talents.where(user: {profile_type: "waiting_for_approval"})
       elsif filter_params[:status] == "Verified"
         talents.where(verified: true)
+      elsif filter_params[:status] == "By Celo Network"
+        chain_id = contract_env == "production" ? Web3::ApiProxy::CELO_CHAIN[2].to_i : Web3::ApiProxy::STAGING_CELO_CHAIN[2].to_i
```
RubenSousaDinis, line 70: I would have a chain_id method to get rid of some duplication in the env conditional.

RubenSousaDinis (commented):
Some small comments. Other than that all good!

RubenSousaDinis (approved):
Well done!

PR #716, opened 2022-10-03
Added task for rewarding polygon launches

MinervaMcG:
## Summary

This PR adds a rake task that can provide 500 TAL as rewards to the first 50 users who launched their token on the polygon network. It also enables the reward during token launch on the polygon network.

## Notion link

https://www.notion.so/talentprotocol/Polygon-Marketing-Launch-f6e6db04739043cfa87d20766415b026


app/services/talents/polygon_launch_reward.rb
```diff
@@ -0,0 +1,22 @@
+require "web3/api_proxy"
```
verdefred, line 1: can we add some tests for this service, please? one of our goals is to have above 90% coverage by the end of the year 💪

verdefred (commented):

verdefred (approved):
Great work!

PR #718, opened 2022-10-06
Add nft image placeholder. Sort tokens by name. Add debounce to updates.

RubenSousaDinis:
## Summary

This PR adds some minor fixes to the profile web3 collections


app/models/erc721_token.rb
```diff
@@ -9,4 +9,30 @@ class Erc721Token < ApplicationRecord
...
+
+  def token_image
+    return token_image_url if token_image_url
+
+    metadata.present? ? ipfs_image(metadata["image"]) : nil
+  end
+
+  private
+
+  def ipfs_image(ipfs_url)
```
verdefred, line 33: can we do a quick test for this method?

verdefred (commented):

app/models/erc721_token.rb
RubenSousaDinis: normally private methods are not tested. But I can do a test for the method calling this one

verdefred (dismissed):

verdefred (approved):

PR #720, opened 2022-10-06
Profile community section

RubenSousaDinis:
## Summary

This PR adds the community section of the profile.

It adds a new model `Connection` that stores the user connection with other users. This model is useful to ease the query and work with pagination and search.


app/models/connection.rb
```diff
@@ -0,0 +1,51 @@
+class Connection < ApplicationRecord
+  belongs_to :user
+  belongs_to :connected_user, foreign_key: :connected_user_id, class_name: "User"
+
+  validates :connected_at, presence: true
+  validates :user_id, uniqueness: {scope: :connected_user_id}
+
+  # Try reading it like this:
```
francisco-leal, line 8: ahah good comment

app/packs/src/components/profile/Community.jsx
```diff
@@ -0,0 +1,402 @@
...
+};
+
+const Community = ({ userId, talent }) => {
+  const { mobile } = useWindowDimensionsHook();
+  const [connections, setConnections] = useState([]);
+  const [pagination, setPagination] = useState({});
+  const { mode } = useTheme();
+  const url = new URL(document.location);
+
+  const options = useMemo(() => {
```
francisco-leal, line 215: Couldn't this be just an array if it never changes based on the input?

```jsx
const options = [
      { name: "All", value: null },
      { name: "Super Connection", value: "super_connection" },
      { name: "Supporter", value: "supporter" },
      { name: "Supporting", value: "supporting" },
      { name: "Follower", value: "follower" },
      { name: "Following", value: "following" },
]
```

francisco-leal (approved):

PR #734, opened 2022-10-16
WIP: Elasticsearch for talents

MinervaMcG:


app/models/tag.rb
```diff
@@ -5,6 +5,9 @@ class Tag < ApplicationRecord
 
   validates :description, presence: true
 
+  # Elasticsearch index update
+  # update_index("talents") { talents }
+
```
verdefred, line 10: if this is not needed, can we delete it?

app/models/goal.rb
```diff
@@ -13,4 +13,14 @@ class Goal < ApplicationRecord
...
+
+  update_index("talents") { career_goal.talent }
+
+  after_save :touch_talent
+
+  private
+
+  def touch_talent
+    career_goal.talent.touch
+  end
```
verdefred, line 25: why is this necessary?

verdefred (commented):
great work, just left a couple of questions

app/models/goal.rb
MinervaMcG: So, the main index that we have is through the talents table, the others are associations with it. When we update the index for talents it looks for the talents updated since the last index update, for this purpose we need to update the updated_at timestamp of talents for it sync successfully with the elasticsearch

config/chewy.yml
```diff
@@ -0,0 +1,3 @@
+# separate environment configs
+test:
+  host: "localhost:9200"
```
verdefred, line 3:
```yml
# separate environment configs
test:
  host: "localhost:9250"
  prefix: "test"
development:
  host: "localhost:9200"
```
can we change this file so it has the development host?

config/initializers/chewy.rb
```diff
@@ -0,0 +1,22 @@
...
+      proc: ->(f) do
+        f.request :aws_sigv4,
+          service: "es",
+          region: ENV["AWS_ELASTICSEARCH_REGION"],
+          access_key_id: ENV["AWS_ELASTICSEARCH_ACCESS_KEY_ID"],
+          secret_access_key: ENV["AWS_ELASTICSEARCH_SECRET_ACCESS_KEY"]
+      end
+    }
+  }
+end
```
verdefred, line 18:
```rb
if Rails.env.production?
  Chewy.settings = {
    host: ENV["AWS_ELASTICSEARCH_URL"],
    port: 443, # 443 for https host
    transport_options: {
      headers: {content_type: "application/json"},
      proc: ->(f) do
        f.request :aws_sigv4,
          service: "es",
          region: ENV["AWS_ELASTICSEARCH_REGION"],
          access_key_id: ENV["AWS_ELASTICSEARCH_ACCESS_KEY_ID"],
          secret_access_key: ENV["AWS_ELASTICSEARCH_SECRET_ACCESS_KEY"]
      end
    }
  }
end

Chewy.settings = {host: "localhost:9200"} if Rails.env.development?
Chewy.settings = {host: "localhost:9250"} if Rails.env.test?
```
can we change to this so we can use it in development? thanks!

PR #757, opened 2022-10-21
Fix dates

francisco-leal:
## Summary

remove inputs that are using dates in favor of dropdowns with months and years.


app/controllers/api/v1/talent/milestones_controller.rb
```diff
@@ -73,7 +73,8 @@ def milestone_params
       :institution,
       :description,
       :link,
-      :category
+      :category,
+      :in_progress
```
RubenSousaDinis, line 77: I believe this should be a computed field in the BE. At least I don't see it being passed in the client side.

RubenSousaDinis (approved):

PR #769, opened 2022-10-27
Remove investor model from the app

RubenSousaDinis:
## Summary

This PR cleans the leftovers from the old profile. The major change is the removal of the investors model.


app/jobs/daily_metrics_job.rb
```diff
@@ -8,7 +8,6 @@ def perform
       total_connected_wallets: total_connected_wallets,
       total_active_users: total_active_users,
       total_dead_accounts: total_dead_accounts,
-      total_talent_profiles: total_talent_profiles,
```
RubenSousaDinis, line 11: All users have a talent profile now.

PR #778, opened 2022-11-06
feat: invites part of partnership or user

shirleyodeng:
## Summary

This PR;
- associates an invite with either a user or a partnership.
- creates an invite for the partnership on creation.

## Notion link

https://talentprotocol.notion.site/Invites-part-of-partnership-or-user-ef5f9f96b8c74734bc4daee05c86e8dd


app/services/invites/create.rb
```diff
@@ -3,33 +3,26 @@ class Create
     MAX_RETRIES = 5
 
     def initialize(params)
-      @user_id = params.fetch(:user_id)
+      @partnership = params.fetch(:partnership, nil)
       @single_use = params.fetch(:single_use, false)
       @talent_invite = params.fetch(:talent_invite, false)
+      @user = params.fetch(:user, nil)
```
shirleyodeng, line 9: Changed this to a record rather than an ID to prevent a double query to the database

RubenSousaDinis (changes requested):
Some comments yo consider above.

We also need to change this piece of code: 

https://github.com/talentprotocol/web-dapp/blob/8edd4007966660b7051f8f7b2556487363617db1/app/blueprints/user_blueprint.rb#L32-L39

```ruby
    field :invited_by do |user, _options|
      if user.invited
        {
          name: user.invited.user.name,
          profile_picture_url: user.invited.user.profile_picture_url
        }
      end
    end
```

app/services/partnerships/create.rb
```diff
@@ -0,0 +1,54 @@
...
+        partnership.logo = Down.open(params["logo_url"])
+      end
+
+      return partnership if partnership.save
+
+      raise CreationError, "Unable to create partnership. Errors: #{partnership.errors.full_messages.join(", ")}"
+    end
+
+    def create_invite!(partnership)
+      Invites::Create.new(partnership: partnership, talent_invite: true).call
```
RubenSousaDinis, line 51: We need to specify the number of invites somehow. It's ok if the info comes from the json file.

lib/tasks/partnerships.rake
```diff
@@ -1,7 +1,7 @@
 namespace :partnerships do
   desc "Create partnership from JSON file"
-  # rake "partnerships:create_from_json[invite_code, spec/fixtures/files/partnerships/utrust.json, 1, 1]"
-  task :create_from_json, [:invite_code, :json_path, :max_uses, :user_id] => :environment do |_t, args|
+  # rake "partnerships:create_from_json[spec/fixtures/files/partnerships/utrust.json]"
+  task :create_from_json, [:json_path] => :environment do |_t, args|
```
RubenSousaDinis, line 4: we still need the max_uses for the invite

shirleyodeng: Fixed! https://github.com/talentprotocol/web-dapp/pull/778/commits/18c8b5495f07327453f57a259d358c33aa6dce46

RubenSousaDinis (approved):
Well done!

PR #798, opened 2022-11-10
Debounce the API request instead of the user input on registration

francisco-leal:
## Summary

When checking the username we must let the user type and never debounce the input field directly. Instead we should debounce the API call.


RubenSousaDinis (approved):
Well done!

PR #804, opened 2022-11-10
Load talents from client and bring back randomness

RubenSousaDinis:
## Summary

This PR makes some changes to the talent page:

- Random talents each day is back
- Load more add new rows to the end
- All filtering is now being done by the backend
- Add admin all view (this makes the All default view similar to all users)


app/packs/src/components/talent/TalentPage.jsx
```diff
@@ -93,38 +106,61 @@ const TalentPage = ({ talents, pagination, isAdminOrModerator, env }) => {
...
-    return desiredTalent;
-  }, [localTalents, watchlistOnly, selectedSort, sortDirection]);
+    setTalents(sortedTalent);
+  }, [selectedSort, sortDirection]);
+
+  const loadTalents = (params) => {
+    setLoading(true);
+    get(`/api/v1/talent?${params.toString()}`).then((response) => {
+      if (response.error) {
+        toast.error(<ToastBody heading="Error!" body={response.error} />);
```
verdefred, line 121: should we also add this to a catch, if the call fails?

app/services/talents/search.rb
```diff
@@ -45,10 +60,18 @@ def filter_by_keyword(talents)
       talents.where(user: users.distinct.select(:id))
     end
 
+    def filter_by_watchlist(talents)
```
verdefred, line 63: can we add a small test that touches on this method?

verdefred (commented):
great work just left a couple of really small comments

app/services/talents/search.rb
RubenSousaDinis: Added!

app/packs/src/components/talent/TalentPage.jsx
RubenSousaDinis: we're not doing that in other parts of the codebase. I prefer to keep it similar

verdefred (approved):

PR #844, opened 2022-12-20
Hide digital collectibles when there's no visible tokens

RubenSousaDinis:
## Summary

This PR hides the digital collectibles section when the user has no visible tokens to display.

It also adds progress status to the goals.

Screenshot 2022-12-20 at 15 49 10


app/blueprints/user_blueprint.rb
```diff
@@ -11,9 +11,18 @@ class UserBlueprint < Blueprinter::Base
...
       user.name
     end
+
     field :is_talent do |user, _options|
       user.talent.present?
     end
+
+    field :is_talent do |user, _options|
+      user.talent.present?
+    end
```
verdefred, line 21: this seems to be repeated

PR #845, opened 2022-12-20
Add WithPersona integration for verifications

verdefred:
## Summary
KYC: Integration with WithPersona

We want to add automatisation to our verification process and, for that, we'll integrate with WithPersona.
All users will now be able to verify themselves with WithPersona and then we'll check if the user's name is the same in the app. If so, the user is automatically verified

## Notion link

https://www.notion.so/talentprotocol/KYC-Integration-with-Persona-6c24b1ed83c94153afacf5c2450407c1
https://www.notion.so/talentprotocol/Name-Fields-on-Profile-25ad8df665274bb997b864f39bd1fec6

## Notes

image


app/packs/src/components/profile/Overview.jsx
```diff
@@ -235,6 +231,45 @@ const Overview = ({
     );
   }, []);
 
+  const verifyTalent = async () => {
+    const client = new Persona.Client({
+      templateId: railsContext.withPersonaTemplateId,
+      environment: railsContext.withPersonaEnvironment,
+      onReady: () => client.open(),
+      onComplete: ({ inquiryId }) => {
+        patch(`/api/v1/with_persona_requests/${withPersonaRequest.id}`);
```
RubenSousaDinis, line 240: Should we have a patch without passing the id? Imagine you open the page at `2022:12:31 23:59:59` and you click verify after 2 seconds. We should update the record in January 2023, right? I believe the BE should get the record using the current date.

RubenSousaDinis: It would also remove the need for the record to be passed to the FE

RubenSousaDinis (commented):
Some comments to discuss

app/packs/src/components/profile/Overview.jsx
verdefred: We would still need to pass how many requests we've made so far, right?

app/packs/src/components/profile/Overview.jsx
```diff
@@ -285,6 +320,8 @@ const Overview = ({
     return talent.profile.headline?.split(" ");
   }, [talent.profile.headline]);
 
+  console.log({ withPersonaRequest });
```
RubenSousaDinis, line 323: remove log

RubenSousaDinis (commented):
Just one log to remove. Would be good to add tests to the job and service in order to not decrease the code coverage.

RubenSousaDinis (commented):
The PR is currently lowering the code coverage. Can we add a couple more tests?

verdefred:
> The PR is currently lowering the code coverage. Can we add a couple more tests?

Yes, definitely! Are you ok with that being done in a separate PR that goes out still today?

app/packs/src/components/profile/edit/EditOverviewModal.jsx
```diff
@@ -487,8 +489,40 @@ const EditOverviewModal = ({ show, hide, talent, setTalent }) => {
...
+          <div className="w-100 mb-5">
+            <TextInput
+              className="mb-2"
+              title="First Name"
+              onChange={(e) =>
+                changeUserAttribute("legalFirstName", e.target.value)
+              }
+              value={editedTalent.user.legalFirstName}
+              error={validationErrors?.legalFirstName}
+              disabled={editedTalent.verified || editedTalent.withPersonaId}
```
RubenSousaDinis, line 502: Should this be disabled only when the user was verified via persona?

RubenSousaDinis (approved):

PR #846, opened 2022-12-21
Fix users destroy

RubenSousaDinis:
## Summary

This PR allows a user to destroy his account.


spec/services/users/destroy_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe Users::Destroy do
+  subject(:destroy_user) { described_class.new(user: user).call }
+
+  let!(:user) { create :user }
```
verdefred, line 6: can we also add a talent with career goals, milestones, etc so we can improve a little bit the coverage?

verdefred (approved):

PR #850, opened 2022-12-21
Change quests based on Talent Mate NFTs

verdefred:
## Summary
Change obsolete information in old quests and add new information regarding the new NFTs

## Notes

image


app/services/tasks/update.rb
```diff
@@ -34,7 +34,8 @@ def give_rewards_for_task(type:, user:)
         user.invites.where(talent_invite: false).update_all(max_uses: nil)
       elsif type == "Quests::VerifiedProfile"
         WhitelistUserJob.perform_later(user_id: user.id, level: "verified")
-        Reward.create!(user: user, amount: 100, category: "quest", reason: "Got verified")
+      elsif type == "Quests::TalentProfile"
+        Reward.create!(user: user, amount: 50, category: "quest", reason: "completed profile")
```
RubenSousaDinis, line 38: We want to do this?

verdefred: I discussed it with @francisco-leal but basically we felt that the Verify quest already had too much things in it and the Talent Profile was giving an obsolete NFT, which didn't make sense anymore. We decided for this approach, knowing that it can backfire and people can farm it. Since we have daily metrics regarding this, we should be on top of it and if we see an anomaly in the metrics we can shut it down

francisco-leal (approved):

PR #854, opened 2022-12-22
Only allow verification for users with a completed profile

verdefred:
## Summary
We're having some verification requests for users with no profile. We want to have a high standard for our profiles and people that can mint our NFTs so we're adding a validation to check if the user has the profile completed before allowing to do the verification process

## Notes

image
image
image
image


app/models/talent.rb
```diff
@@ -78,6 +78,10 @@ def profile_picture_url
     profile_picture(:default)&.url || super
   end
 
+  def profile_completed?
```
RubenSousaDinis, line 81: I think this method belongs to the user model. We already have a similar one there: https://github.com/talentprotocol/web-dapp/blob/bd1e4514387b2a0d2440f49399d9ed9c9bd22904/app/models/user.rb#L123-L125

RubenSousaDinis (commented):
Just one comment

app/models/talent.rb
verdefred: Good catch, I got distracted with the scope profile_complete in here. I'll make the change

RubenSousaDinis (approved):
Well done!

app/packs/src/components/profile/Overview.jsx
```diff
@@ -553,16 +553,34 @@ const Overview = ({
...
-                              onClick={() => verifyTalent()}
-                            />
-                          )}
+                        {!talent.verified && (
+                          <Button
+                            className="mr-2"
+                            type="primary-default"
+                            onClick={() => verifyTalent()}
+                            disabled={
+                              !talent.profileCompleted || talent.withPersonaId
```
RubenSousaDinis, line 562: should this be `user.profileCompleted`?

RubenSousaDinis (approved):

PR #856, opened 2022-12-22
Add persona failed notifications

verdefred:
## Notes

image
image
image


config/locales/en.yml
```diff
@@ -62,3 +62,9 @@ en:
     talent_application_rejected_notification:
       title: "Talent application rejected"
       body: "Your talent application has been rejected by %{name}"
+    user_names_verification_failed_notification:
+      title: "Something's wrong 🤔"
```
RubenSousaDinis, line 66: No mention to verification in the app notification?

verdefred: I think you're right. I've talked with Sara and we'll change to "Verification names don't match"

app/services/with_persona/failed_inquiry.rb
```diff
@@ -8,7 +8,12 @@ def initialize(talent:)
 
     def call
       talent.update(with_persona_id: nil)
-      # failed notification
+      CreateNotification.new.call(
```
RubenSousaDinis, line 11: Test the notification service is called?

app/services/with_persona/approve_inquiry.rb
```diff
@@ -14,7 +14,12 @@ def call
         Tasks::Update.new.call(type: "Tasks::Verified", user: user)
       else
         talent.update(with_persona_id: nil)
-        # failed notification
+        CreateNotification.new.call(
```
RubenSousaDinis, line 17: Test the notification service is called?

RubenSousaDinis (approved):

PR #876, opened 2023-01-05
feat: added 5 different avatars by default

helloiambguedes:
## Summary

Added 5 different avatars style by default

## Notion link

[](https://www.notion.so/talentprotocol/Default-Avatar-with-5-colors-b427fa5846f94f73bfb459f5e498c94b)

## How to test?

Check users default images

## Notes

N/A


app/packs/src/components/talent/TalentProfilePicture.jsx
```diff
@@ -17,16 +16,15 @@ const TalentProfilePicture = ({
...
   const { mode } = useTheme();
 
   const imgSrc = () => {
     if (src) {
       return src;
-    } else if (mode() === "light" && !straight) {
-      return DefaultProfilePicture;
-    } else if (mode() === "dark" && !straight) {
-      return DefaultProfilePicture;
+    } else if ((mode() === "light" || mode() === "dark") && !straight) {
```
verdefred, line 26: I think we can remove the mode here, since it doesn't matter. it would be:
`} else if (!straight) {`

helloiambguedes: That's true, will change 👌

francisco-leal (dismissed):

verdefred (approved):

PR #884, opened 2023-01-11
Fix elasticsearch index and search service

verdefred:
## Summary
* Add missing fields to index
* talent_token -> token
* Add filter by watchlist
* Add filter by discovery row
* Fix query for status


app/indexes/talents_index.rb
```diff
@@ -1,16 +1,40 @@
...
+    end
+  end
+
+  field :token, value: ->(talent) { talent.talent_token } do
+    field :ticker, :contract_id, :chain_id
+    field :deployed_at, type: "date"
   end
 
-  field :talent_token do
+  field :token, value: ->(talent) { talent.talent_token } do
```
RubenSousaDinis, line 37: Not a huge fan of the naming here. `token` is too generic. It can be an NFT or a ERC20 token and we already have that naming in a concern. I would keep talent_token in the codebase to refer to the tokens launched by the user.

verdefred: Currently we have the name "token" referring to the talent token all over our FE. if you go to our TalentBlueprint you can see that we have given the name "token" to the talent token association:
`association :talent_token, blueprint: TalentTokenBlueprint, view: :normal, name: :token`
I kept the naming here so it's consistent in the app

verdefred: I can change the FE in this PR, then 👍

RubenSousaDinis: Great! Thanks for that

RubenSousaDinis (approved):
Well done!

PR #906, opened 2023-01-31
feat: support page for moderators

helloiambguedes:
## Summary

Added a support page for moderators where they can search and check data about the users

## Notion link

[Task](https://www.notion.so/talentprotocol/Creating-a-tool-or-recreating-a-role-to-see-more-info-on-user-profile-to-improve-support-experience-ebd333bc3c6c43d4a6cdf9744f6a8b28)

## How to test?

Access `/support`

## Notes

N/A


app/controllers/support_controller.rb
```diff
@@ -0,0 +1,28 @@
+class SupportController < ApplicationController
+
+  def index
+  end
+
+  def user_page
+    user_ref = User.
```
francisco-leal, line 7: You can use

```rb
User.find(params(:id) # will throw an error if it doesn't find anyone (which already happens on your code given that you do user_ref[0] on line 10.
User.find_by(id: params[:id]) # which will return nil if it doesn't find any user
```

app/controllers/support_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      where(id: params[:id]);
+    @user = UserBlueprint.render_as_json(user_ref, view: :with_pictures)[0];
+    @userData = user_ref[0];
+    @supportings = TalentSupporter.where(supporter_wallet_id: @userData.wallet_id);
+    @supporteds = User.joins(:talent)
+      .joins("JOIN talent_tokens ON talent_tokens.talent_id = talent.id")
+      .joins("JOIN talent_supporters ON talent_supporters.talent_contract_id = talent_tokens.contract_id")
+      .where("talent_supporters.supporter_wallet_id = '#{@userData.wallet_id}'");
```
francisco-leal, line 15: `.where(talent_supporters: { supporter_wallet_id: @userData.wallet_id})`

app/controllers/support_controller.rb
```diff
@@ -0,0 +1,28 @@
+class SupportController < ApplicationController
+
+  def index
+  end
+
+  def user_page
+    user_ref = User.
+      where(id: params[:id]);
+    @user = UserBlueprint.render_as_json(user_ref, view: :with_pictures)[0];
+    @userData = user_ref[0];
```
francisco-leal, line 10: This isn't really the userData, this is the user itself. I would refactor this to merge with my suggestion above:
```rb
def user_page
  @user = User.find_by(id: params[:id))
  @userData = UserBlueprint.render_as_json(@user, view: :with_pictures); # the blueprint is smart, so since you're only passing one object, it will only return one value
  #...
end
```

app/controllers/support_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    @supportings = TalentSupporter.where(supporter_wallet_id: @userData.wallet_id);
+    @supporteds = User.joins(:talent)
+      .joins("JOIN talent_tokens ON talent_tokens.talent_id = talent.id")
+      .joins("JOIN talent_supporters ON talent_supporters.talent_contract_id = talent_tokens.contract_id")
+      .where("talent_supporters.supporter_wallet_id = '#{@userData.wallet_id}'");
+    @totalRewards = Reward
+      .where(user_id: 2)
```
francisco-leal, line 17: You probably don't want this hardcoded value

config/routes.rb
```diff
@@ -9,6 +9,17 @@
...
 
+  # Moderator area
+  constraints Clearance::Constraints::SignedIn.new { |user| user.moderator? } do
+    # support_controller.rb
+    # views/controller/index.html.erb
+    # -> GET api/v1/support (:username, :wallet/address)
+    get "/support" => "support#index"
+    get "/support/search/" => "support#search"
+    get "/support/user/:id" => "support#user_page"
+  end
```
francisco-leal, line 20: This would be a more ruby way of doing these routes so that it allows for us to extend the 
```rb
resources :support, only: [:index] do # resources defines all the normal REST routes to access a resource, but we're adding a only index modifier
  get :search, to: "support#search"
  resources :users, only: [:show] # same as above but we only want the user/:id GET route
end
```

francisco-leal (approved):

app/packs/src/components-v2/support/index.jsx
```diff
@@ -0,0 +1,27 @@
+import React, { useMemo } from "react";
+import { Input } from "src/components-v2/shared-styles/input";
+import { Avatar } from "src/components-v2/shared-styles/avatar";
+import { useSupportToolUtils } from "./hooks/use-support-tool-utils";
+import { Container, ListContainer, ProfileContainer } from "./styles";
+
+export const Support = () => {
+    const utils = useSupportToolUtils();
+    const usersList = useMemo(() => 
+        utils.users.map(user => (
```
RubenSousaDinis, line 10: 4 spaces instead of 2?

config/routes.rb
```diff
@@ -9,6 +9,17 @@
   end
   # end Admin
 
+  # Moderator area
+  constraints Clearance::Constraints::SignedIn.new { |user| user.moderator? } do
```
RubenSousaDinis, line 13: Admins should also be able to visit this page.

RubenSousaDinis (changes requested):
Some changes are still needed. Also left some comments to discuss

app/controllers/support_controller.rb
```diff
@@ -0,0 +1,22 @@
+class SupportController < ApplicationController
+  def index
+  end
+
+  def user_page
+    @user = User.find(params[:id])
+    @user_with_pictures = UserBlueprint.render_as_json(@user, view: :with_pictures)
+    @supportings = TalentSupporter.where(supporter_wallet_id: @user.wallet_id)
+    @supporteds = @user.portfolio
```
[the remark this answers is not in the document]
helloiambguedes: Was not in the Notion task, is it worth it?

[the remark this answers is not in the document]
RubenSousaDinis: I ended up understanding why you need the two sets of data.

RubenSousaDinis (approved):

PR #937, opened 2023-02-16
fix: fixed some ui issues and chat order glitch

helloiambguedes:


app/packs/src/components/chat/Chat.jsx
```diff
@@ -67,9 +67,6 @@ const Chat = ({ chats, user, pagination }) => {
       setMessengerUsername(response.username);
       setLastOnline(response.lastOnline);
       setGettingMessages(false);
-      if (response.readChat) {
```
RubenSousaDinis, line 70: Without this a user in the second page will not be displayed in the sidebar

RubenSousaDinis: Not sure if it's ok or not

RubenSousaDinis (approved):

PR #964, opened 2023-03-06
feat: new onboarding

helloiambguedes:
## Summary

New OnBoarding experience by the homepage scope

## Notion link

[](https://www.notion.so/talentprotocol/Implement-Onboarding-Experience-d333ceeb6b514fcda3c7d60a28b6842d?pvs=4)

## How to test?

Create a new account and login :)


app/views/onboard/sign_up.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= react_component("SignUpPage",
+    props: {
+      linkedinClientId: ENV["LINKEDIN_CLIENT_ID"],
+      linkedinRedirectUri: ENV["LINKEDIN_REDIRECT_URI"],
+      step: 1,
+      captchaKey: ENV["RECAPTCHA_SITE_KEY"],
+      inviteCode: @invite&.code,
+      profilePictureUrl: @user&.profile_picture_url,
+      name: @user&.display_name || @user&.username,
```
RubenSousaDinis, line 9:
```suggestion
      name: @user&.name
```

app/views/onboard/index.html.erb
```diff
@@ -0,0 +1 @@
+<%= react_component("OnBoardPage") %>
```
RubenSousaDinis, line 1: missing empty line

app/views/onboard/forgot_password.erb
```diff
@@ -0,0 +1 @@
+<h1>Forgot Password</h1>
```
RubenSousaDinis, line 1: missing empty line

app/packs/src/components-v2/sign-up-flow/welcome-footer/styled.ts
```diff
@@ -0,0 +1,11 @@
...
+import { buildColor } from "@talentprotocol/design-system";
+
+export const FirstStepActionArea = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  gap: 8px;
+  padding: 16px;
+  border-top: 1px solid ${buildColor("surfaceHover01")};
+`;
```
RubenSousaDinis, line 11: missing empty line

RubenSousaDinis (commented):
Added some comments and missing tests

app/controllers/sessions_controller.rb
```diff
@@ -29,34 +29,6 @@ def destroy
...
-
-      return render "profiles/show"
-    end
-
-    render template: "sessions/new"
-  end
-
   private
 
   def profile_subdomain?
```
RubenSousaDinis, line 34: delete private area, not being used

spec/services/users/create_spec.rb
```diff
@@ -11,7 +11,12 @@
       password: password,
       username: username,
       invite_code: invite_code,
-      wallet_id: wallet_id
+      wallet_id: wallet_id,
+      gender: gender,
+      nationality: nationality,
+      location: location,
+      career_needs: ["Mentoring others"],
+      tags: ["Web3"]
```
RubenSousaDinis, line 19: Adding this here is not really testing the new data provided. It would be good to assert that the params passed are being stored. We can use the existing test here for that: https://github.com/talentprotocol/web-dapp/pull/964/files#diff-39f0cd501c91821665dc7f2122f73ab9e4447fda39f586431a2107f75ff3402aR43

app/controllers/onboard_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      )
+
+      @with_persona_request = WithPersonaRequestBlueprint.render_as_json(WithPersonaRequest.current_month_persona_request)
+
+      render "profiles/show"
+    end
+  end
+
+  def sign_up
+    if params[:invite_code].present?
```
RubenSousaDinis, line 29: We have removed the behaviour when the user was sign in and accessed this page:

```
    if current_user.present?
      return redirect_to user_path(username: params[:invite_code]) if user
      return redirect_to talent_index_path
    end
```

RubenSousaDinis (commented):
Added some comments

app/controllers/onboard_controller.rb
```diff
@@ -0,0 +1,41 @@
+class OnboardController < ApplicationController
+  def sign_in
+    if profile_subdomain?
```
verdefred, line 3: what happens if we don't have a TAL_BASE_DOMAIN?

app/packs/src/components-v2/sign-up-flow/utils/field-validators.js
```diff
@@ -0,0 +1,13 @@
...
+  callback(result);
+  return result;
+};
+
+export const validatePassword = (password, callback) => {
+  const regex = /^(?=.*\d)(?=.*[a-z])(?=.*[A-Z]).{8,}$/;
+  const result = regex.test(password);
+  callback(result);
+  return result;
+};
```
verdefred, line 13: I think this validators can be used in other parts of our app so I think it makes sense to put them in a utils folder outside of the sign-up-flow. what do you think?

app/packs/src/utils/window.js
```diff
@@ -8,10 +8,10 @@ const getWindowDimensions = () => {
   };
 };
 
-const useWindowDimensionsHook = () => {
+const useWindowDimensionsHook = (definedWidth = 992) => {
   const [windowDimensions, setWindowDimensions] = useState(getWindowDimensions());
 
-  const mobile = windowDimensions.width < 992;
+  const mobile = windowDimensions.width < definedWidth;
```
verdefred, line 14: I don't understand this change. If I define a different width, like 1440, then it will be mobile if it's smaller than 1440? Why do we need to pass a width?

app/controllers/onboard_controller.rb
helloiambguedes: It should render the sign_in page normally
This should just cover users going to TAL_BASE_DOMAIN pages

app/packs/src/components-v2/sign-up-flow/utils/field-validators.js
helloiambguedes: That's something I plan to add to the ecosystem repo.
Just felt like creating a new NPM package would be overkill.

As @francisco-leal is picking up the new architecture migration soon I didn't thought it would be worth it

app/packs/src/utils/window.js
helloiambguedes: I added this because the new mobile breakpoint is `1240`. That's something we can discuss with PP.
As I don't want to touch the old stuff but wanted to still reuse this for now (while I don't have the utils lib ready) I did this quick trick.

verdefred: Got it, thanks for the explanation. I missed the breakpoint changing in the design system

verdefred (approved):

PR #968, opened 2023-03-07
Add endpoint to fetch a list of talents

RubenSousaDinis:
## Summary

This PR adds an endpoint to retrieve a list of talents.


verdefred (approved):

app/controllers/api/v1/public_api/talents_controller.rb
```diff
@@ -9,9 +9,41 @@ def show
...
 
+  def index
+    users = User.all
+    users = users.where("wallet_id IN (:ids) OR username IN (:ids)", ids: filter_params[:ids]) if filter_by_ids?
+
+    pagy, page_users = pagy_uuid_cursor(
+      users,
+      before: cursor,
+      items: per_page,
+      order: {created_at: :desc, uuid: :desc}
```
verdefred, line 20: can we order just by uuid? I'm saying this because that way we would have O(limit) instead of O(2*N-W)
image
https://medium.com/swlh/how-to-implement-cursor-pagination-like-a-pro-513140b65f32

RubenSousaDinis: The uuid does not give the results sorted, since it's a random id being generated. Our API is returning all sets sorted by `created_at` to be consistent in the API designs. Our consumers should expect to always see the latests records when they query our endpoints. We can change this if we ended up having performance issues, but from the tests I did the endpoints are responding fast as is.

I can probably add an index to the created_at field tho. But I can do that in another PR

verdefred: Why don't we order just by `created_at`?

verdefred (approved):

PR #971, opened 2023-03-08
Fetch current user data on rewards and profiles pages

RubenSousaDinis:
## Summary

This PR keeps replacing the propagation of the current user data from the server to the client.


app/packs/src/components/design_system/progress_circle/index.jsx
```diff
@@ -16,13 +16,13 @@ const ProgressCircle = ({ id, width, progress, completedTasks, allTasks, done =
   };
 
   useEffect(() => {
-    if (document.getElementById(`circle-${id}`)?.querySelector(".progress-ring")) {
-      setProgress(document.getElementById(`circle-${id}`).querySelector(".progress-ring"), progress);
+    if (document.getElementById(`circle-${type}`)?.querySelector(".progress-ring")) {
+      setProgress(document.getElementById(`circle-${type}`).querySelector(".progress-ring"), progress);
     }
-  }, [document.getElementById(`circle-${id}`)?.querySelector(".progress-ring")]);
+  }, [document.getElementById(`circle-${type}`)?.querySelector(".progress-ring")]);
```
helloiambguedes, line 22: This is an anti-pattern, we should use refs for these things

app/packs/src/components/rewards/index.jsx
```diff
@@ -43,14 +43,21 @@ const Rewards = ({
...
   const [localInvitedUsers, setLocalInvitedUsers] = useState(invitedUsers);
   const [localPagination, setLocalPagination] = useState(pagination);
 
+  const { currentUser, fetchCurrentUser } = loggedInUserStore();
+
+  useEffect(() => {
+    if (!currentUser) {
+      fetchCurrentUser();
+    }
+  }, []);
```
helloiambguedes, line 59: This might throw an eslint error because the `currentUser` is a dependency of the `useEffect`

app/blueprints/user_blueprint.rb
```diff
@@ -1,12 +1,12 @@
 class UserBlueprint < Blueprinter::Base
   fields :id, :uuid
 
-  view :basic_with_pictures do
+  view :support do
```
helloiambguedes, line 4: Not PR related but I believe we should aim for these kind of utilities to be available for HTTP calls.
If we want to detach the BE from the FE we should aim to remove feature related logic and have more of a SAAS mindset.

Just a thought

helloiambguedes (dismissed):

app/blueprints/user_blueprint.rb
RubenSousaDinis: That's what I'm working on at the moment eheh

app/packs/src/components/design_system/progress_circle/index.jsx
RubenSousaDinis: but the problem was already there, right?

app/packs/src/components/rewards/index.jsx
RubenSousaDinis: I can simply remove the if since it's not doing much. The currentUser will always need to be fetched in the component load

PR #977, opened 2023-03-10
Add whitelist action to approval process

verdefred:
## Notes

image
image
image


app/packs/src/components/profile/ApprovalConfirmationModal.jsx
```diff
@@ -38,6 +46,55 @@ const ApprovalConfirmationModal = ({ show, hide, talent, setTalent }) => {
...
+
+    setFirstLoading(false);
+  });
+
+  useEffect(() => {
+    setupChain();
+  }, []);
+
+  if (firstLoading) {
+    return <></>;
```
francisco-leal, line 95: This could probably be a spinner, or just a warning message. If we don't add that let people know it might require a bit of loading to pull our the information

app/packs/src/onchain/index.js
```diff
@@ -221,6 +223,33 @@ class OnChain {
     }
   }
 
+  async loadFactoryV3() {
```
francisco-leal, line 226: Shouldn't this replace the existing loadFactory function?

verdefred: Yes, I think you're right. I'll replace the existing one with this one

PR #988, opened 2023-03-14
feat: forgot-password revamp

helloiambguedes:
## Summary

New forgot password experience

## Notion link

[](https://www.notion.so/talentprotocol/Revamp-Forgot-Password-Experience-e33c3960f90f400495ffc3d8ca352b96?pvs=4)

## How to test?

This must be tested in dev as we cannot fully do it locally.

## Notes

Let's sync with the virtual wallet release as this may clog dev for some hours if something fails


app/views/onboard/reset_password.erb
```diff
@@ -0,0 +1,7 @@
+<%= react_component("SetPasswordPage",
+    props: {
+        token: params[:token],
```
RubenSousaDinis, line 3: This is an example of something we should avoid going forward if we want to decouple the BE and FE. Also, the token and user_id are coming from the URL, right? We can access them in the client by fetching them from the page URL?

app/views/onboard/forgot_password.erb
```diff
@@ -0,0 +1,4 @@
+<%= react_component("ForgotPasswordPage",
```
RubenSousaDinis, line 1: File name is missing `.html`

app/views/onboard/reset_password.erb
```diff
@@ -0,0 +1,7 @@
+<%= react_component("SetPasswordPage",
```
RubenSousaDinis, line 1: File name is missing `.html`

RubenSousaDinis (commented):
Added some comments

app/views/onboard/reset_password.erb
helloiambguedes: We do as I didn't add the url cleaning logic.
Will change.

RubenSousaDinis (approved):

PR #1043, opened 2023-04-04
Fetch tal domain profile correctly

RubenSousaDinis:
## Summary

This PR brings back tal domains


helloiambguedes (approved):

app/packs/src/components/profile/show.jsx
```diff
@@ -44,6 +44,10 @@ const Show = ({ railsContext, withPersonaRequest, profileSubdomain }) => {
   }, []);
 
   useEffect(() => {
+    if (profileSubdomain) {
```
verdefred, line 47: should we add the `profileSubdomain` to the dependency array? not sure where this changes

RubenSousaDinis: This is still coming from the server side

verdefred (approved):

PR #1152, opened 2023-05-08
feat: notifications panel update

shirleyodeng:
## Summary

This PR updates the notifications panel.

Screenshot 2023-05-08 at 10 16 08

Screenshot 2023-05-08 at 09 26 44

Screenshot 2023-05-08 at 08 46 24

Screenshot 2023-05-08 at 09 33 38

Screenshot 2023-05-08 at 09 33 23

## Notion link

https://talentprotocol.notion.site/Bounty-Notifications-Panel-Update-cdcb8d5bd67e40269698e623ec1cee00


config/locales/en.yml
```diff
@@ -32,51 +32,20 @@
...
-    talent_application_rejected_notification:
-      title: "Talent application rejected"
-      body: "Your talent application has been rejected by %{name}"
-    user_names_verification_failed_notification:
-      title: "Verification names don't match 🤔"
-      body: "Added names don't match. Please check again!"
-    user_persona_verification_failed_notification:
-      title: "Verification failed 😭"
-      body: "Make sure documents are legible to submit them"
+      body: "Staked XXX USDC/cUSD to endorse you."
```
shirleyodeng, line 50: @RubenSousaDinis, what's the best way to get the stake value? I see the notification is created here and we can pass the value in the extra_params but can I access the value in this class?

https://github.com/talentprotocol/web-dapp/blob/6ad0b5f74b96821ee2be0ff51a168ef4a35b1b39/app/services/stakes/create.rb#L17-L22

RubenSousaDinis: We need to start passing in from the client here: https://github.com/talentprotocol/web-dapp/blob/112e06fe0a7f704dd59817f7862aba8f899d6d12/app/packs/src/components/token/StakeModal.jsx#L145-L148.

app/controllers/api/v1/notifications_controller.rb
```diff
@@ -22,17 +22,16 @@ def index
   end
 
   def mark_all_as_read
-    current_user.notifications.where(read_at: nil).update_all(read_at: Time.current)
+    current_user.notifications.in_app.where(read_at: nil).update_all(read_at: Time.current)
```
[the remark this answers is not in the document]
shirleyodeng: No reason, could be either as in the other notifications are just for emails and won't be used on the client side but I can mark all 👍

config/locales/en.yml
shirleyodeng: @RubenSousaDinis, added here: https://github.com/talentprotocol/web-dapp/pull/1152/commits/e20080db43270a905bfd98a76ef1ef9c824067b5

app/packs/src/components/token/StakeModal.jsx
```diff
@@ -142,9 +142,7 @@ const StakeModal = ({
       const _tokenAvailability = await chainData.getTokenAvailability(targetToken, true);
       setMaxMinting(_tokenAvailability);
 
-      await post(`/api/v1/stakes`, {
-        stake: { token_id: tokenId }
-      }).catch(e => console.log(e));
+      await post(`/api/v1/stakes`, { stake: { amount: amount * 10, token_id: tokenId } }).catch(e => console.log(e));
```
shirleyodeng, line 145: @RubenSousaDinis is it right to multiple amount by 10? I can see that this is done below:

https://github.com/talentprotocol/web-dapp/blob/e20080db43270a905bfd98a76ef1ef9c824067b5/app/packs/src/components/token/StakeModal.jsx#L251

app/packs/src/components/talent/Edit/Settings.jsx
```diff
@@ -404,7 +404,7 @@ const Settings = props => {
           <P2
             className="w-100 text-left"
             mode={mode}
-            text="For each type of notification you can select to receive an immediate email notification, a daily email digest or to not receive any email."
+            text="For each type of notification you can select to receive an immediate email notification or to not receive any email."
```
shirleyodeng, line 407: Screenshot 2023-05-11 at 23 25 30

app/packs/src/components/design_system/notification/index.jsx
```diff
@@ -1,37 +1,74 @@
-import React from "react";
-import { Wallet, Rocket, Chat, Talent, Star, Check, Globe, Reward, Help } from "src/components/icons";
-import { P3 } from "src/components/design_system/typography";
-import { lightTextPrimary03, darkTextPrimary03 } from "src/utils/colors";
+import { Button } from "@talentprotocol/design-system";
 import cx from "classnames";
+import React from "react";
+import { P2, P3 } from "src/components/design_system/typography";
```
helloiambguedes, line 4: We can use `@talentprotocol/design-system` with the most updated layout guidelines

app/packs/src/components/notifications/index.jsx
```diff
@@ -83,21 +62,36 @@ const Notifications = ({ mode }) => {
...
     }
-    if (notification.url) {
+  };
+
+  const onNotificationClick = async (notification, action) => {
+    if (!notification.read) await put(`/api/v1/notifications/${notification.id}/mark_as_read`);
+
+    if (action) {
+      makeActionRequest(action).then(() => {
+        loadNotifications();
```
shirleyodeng, line 92: @RubenSousaDinis - after making the action request.. I'm not sure what to do here so I've just re-loaded the notifications but this will set it to be first page of notifications.. as I'm not sure we'll know which page the notification that was action was on.. if that makes sense. Otherwise, I can close the notifications dropdown after making the request?

RubenSousaDinis (approved):
Great work! Thanks

PR #1153, opened 2023-05-08
feat: new homepage

helloiambguedes:


app/controllers/api/v1/public_api/activity_controller.rb
```diff
@@ -0,0 +1,7 @@
+class API::V1::PublicAPI::ActivityController < API::V1::PublicAPI::APIController
+  def index
+    activity = Activity.order(created_at: :desc).limit(8)
```
RubenSousaDinis, line 3: There's no pagination here? What if I want to see more activities? We need to make this endpoints ready to be consumed by other apps as well. All endpoints under `API::V1::PublicAPI` should be ready to be consumed by external parties. 

Also, the variable should be called activities since there's more than one activity being serialized.

app/controllers/api/v1/public_api/career_updates_controller.rb
```diff
@@ -46,6 +46,8 @@ def create
       career_update: API::CareerUpdateBlueprint.render_as_json(career_update, view: :normal)
     }
 
+    ActivityIngestJob.perform_later("career_update", JSON.generate(response_body[:career_update]), sender.id, nil)
```
RubenSousaDinis, line 49: We should avoid passing nil explicitly. We can make the variable nil by default on the job

app/controllers/api/v1/public_api/leaderboard_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+        end
+      end
+    talents = User.where(id: results.keys)
+    talents.each do |v|
+      parsed_count_hash[v[:uuid]] = count_hash[v[:id]][:count]
+    end
+    render json: {
+      results: parsed_count_hash,
+      users: API::TalentBlueprint.render_as_json(talents, view: :normal)
+    }, status: :ok
```
RubenSousaDinis, line 29: Our Public API follows a pattern that's not being respected here. We use the following:

```
{
   <collection_name>: [ ],
   pagination: {}
}
```

Pagination object is optional.

```suggestion
    render json: {
      leaderboard: {
        results: parsed_count_hash,
        users: API::TalentBlueprint.render_as_json(talents, view: :normal)
        }
    }, status: :ok
```

RubenSousaDinis: Check an example here: https://docs.talentprotocol.com/api/api-reference/talents

app/controllers/api/v1/public_api/talents_controller.rb
```diff
@@ -33,6 +33,13 @@ def index
     render json: response_body, status: :ok
   end
 
+  def recommended
+    subscribed_users = Subscription.where(user_id: current_user.id)
+    recomended_users = User.where.not(id: subscribed_users.select(:subscriber_id)).select("setseed(0.#{Date.today.jd}), *").order("RANDOM()").limit(3)
+    response_body = API::TalentBlueprint.render_as_json(recomended_users, view: :normal)
+    render json: response_body, status: :ok
```
RubenSousaDinis, line 40: Same comment here. API response pattern not being respected

app/models/user.rb
```diff
@@ -50,6 +50,10 @@ class User < ApplicationRecord
   has_many :erc20_tokens, dependent: :destroy
   has_many :erc721_tokens, dependent: :destroy
 
+  # Activities
+  has_many :origin_activities, class_name: "Activity", foreign_key: "origin_user_id"
```
RubenSousaDinis, line 54: missing `dependent: :destroy` on both

RubenSousaDinis (commented):
Add some initial comments.

app/controllers/api/v1/public_api/leaderboard_controller.rb
```diff
@@ -0,0 +1,31 @@
+class API::V1::PublicAPI::LeaderboardController < API::V1::PublicAPI::APIController
+  def index
+    current_race = Race.active_race
+    count_hash = {}
+    parsed_count_hash = {}
+    results = User
```
francisco-leal, line 6: I think this endpoint is going to break if we get a lot of inbound users. Let's spend 10m working on it

app/controllers/api/v1/users_controller.rb
```diff
@@ -1,8 +1,8 @@
 class API::V1::UsersController < ApplicationController
   def index
     @users = search_params[:name].present? ? filtered_users : filtered_users.limit(20)
-
-    render json: {
+    
+    return render json: {
```
francisco-leal, line 5: why do you need this return?

app/packs/src/components-v2/my-wallet-widget/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+  Card,
+  CardContainer,
+  Container,
+  FooterActions,
+  RecentActivityContainer,
+  TitleContainer
+} from "./styled";
+import { Avatar, Button, TextLink, Typography } from "@talentprotocol/design-system";
+
+export const MyWalletWidget = () => {
```
francisco-leal, line 14: We're not using this widget right? (asking because it has hardcoded values

app/packs/src/components-v2/quests-widget/index.jsx
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+import { Container, QuestData, QuestEntry, QuestReward, RewardTag } from "./styled";
+import { Icon, Typography } from "@talentprotocol/design-system";
+
+export const QuestsWidget = () => {
```
francisco-leal, line 5: Same here, just double checking that we're not using this widget.

app/packs/src/components-v2/quick-navigator/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+        </Typography>
+      </ButtonContainer>
+      <ButtonContainer>
+        <Button hierarchy="secondary" size="medium" leftIcon="chat" iconColor="primary" href="/messages" />
+        <Typography specs={{ variant: "p3", type: "medium" }} color="primary01">
+          Messages
+        </Typography>
+      </ButtonContainer>
+      <ButtonContainer>
+        <Button hierarchy="secondary" size="medium" leftIcon="wallet" iconColor="primaryDisable" isDisabled />
```
francisco-leal, line 27: Why is the wallet disabled? We can redirect to the portfolio for now (and call it portfolio)

app/packs/src/components-v2/recommended-teams-widget/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+import talentHouseImg from "./assets/talent-house.png";
+
+export const RecommendedTeamsWidget = ({}) => {
+  const [members, setMembers] = useState({
+    top100: [],
+    talentHouse: [],
+    laMansineta: []
+  });
+  useEffect(() => {
+    Promise.all([axios.get("/api/v1/talent?page=1&discovery_row_id=15"), axios.get("/api/v1/talent?page=1&discovery_row_id=6"), axios.get("/api/v1/talent?page=1&discovery_row_id=12")])
```
francisco-leal, line 25: These values are hardcoded, we need to make sure they work in dev and production

app/views/homepage/index.html.erb
```diff
@@ -0,0 +1 @@
+<%= react_component("HomepagePage", prerender: false) %>
```
francisco-leal, line 1: missing empty line

config/routes.rb
```diff
@@ -32,7 +32,7 @@
 
   # Business - require log-in
   constraints Clearance::Constraints::SignedIn.new { |user| user&.onboarding_complete? } do
-    root to: "discovery#index", as: :user_root
+    root to: "homepage#index", as: :user_root
```
francisco-leal, line 35: In order to help us track this feature and it's usage weren't we going to use `/home`?

app/packs/src/components-v2/my-wallet-widget/index.jsx
helloiambguedes: The UI is ready for the virtual wallet data but we are not using it already

app/packs/src/components-v2/quests-widget/index.jsx
helloiambguedes: Same thing as the above but for the quests bounty.

app/packs/src/components-v2/quick-navigator/index.jsx
```diff
@@ -0,0 +1,64 @@
+import React from "react";
+import { ButtonContainer, Container } from "./styled";
+import { Button, Typography } from "@talentprotocol/design-system";
+
+export const QuickNavigator = ({ username }) => {
+  return (
+    <Container>
+      <ButtonContainer>
+        <Button hierarchy="secondary" size="medium" leftIcon="search" iconColor="primary" isDisabled />
```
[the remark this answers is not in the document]
helloiambguedes: I've postponed this as I didn't want to call the search modal at the time (it would give a bit of boring work) and kinda forgot about it after 😂

app/packs/src/components-v2/recommended-teams-widget/index.jsx
helloiambguedes: They will just work in prod
I think we should refactor this as we don't have a way to explicitly search for collections in the API

config/routes.rb
helloiambguedes: Will do that today

app/controllers/api/v1/public_api/leaderboard_controller.rb
helloiambguedes: I'm missing something on the pagination here as well

app/controllers/api/v1/users_controller.rb
helloiambguedes: I don't

PR #1172, opened 2023-05-15
build: added code-splitting 

helloiambguedes:


config/webpack/base.js
```diff
@@ -23,18 +25,8 @@ const customConfig = {
...
-      cacheGroups: {
-        commons: {
-          test: /[\\/]node_modules[\\/]/,
-          name: "vendor",
-          chunks: "all"
-        }
-      }
-    }
-  }
+    // new TerserPlugin()
```
verdefred, line 28: should this be commented?

helloiambguedes: I left this here as a reference for what I have in mind to do in the future.
We already have this ability just want to make it a little bit better but React on Rails is not cooperating very much

francisco-leal (approved):

PR #1176, opened 2023-05-15
New quests and user quests

RubenSousaDinis:
## Summary

This PR adds two new models:

- v2_quests (our new quests model)
- user_v2_quests (stores the relationship between a user and a quest and the amount credited)

## Notion link

https://www.notion.so/talentprotocol/Earn-Page-Implementation-7e9e213bc9a04f3a90d11fc2f4284980?pvs=4#32da895b9dce4ae199c96aff20a7a402


app/models/v2_quest.rb
```diff
@@ -0,0 +1,4 @@
+class V2Quest < ApplicationRecord
```
verdefred, line 1: it's a bit of a preference style but I would put the version last, so it reads QuestV2 (also easier to find)

RubenSousaDinis: The plan is to call it `Quest` in the last PR. So not sure it's worth changing it

verdefred: got it 💪

verdefred (approved):

PR #1188, opened 2023-05-16
ref: homepage improvements

helloiambguedes:


app/controllers/api/v1/talent_controller.rb
```diff
@@ -16,7 +16,8 @@ def index
       talents: talents,
       pagination: {
         currentPage: paging[:current_page],
-        lastPage: paging[:last_page]
+        lastPage: paging[:last_page],
+        total: talents.count,
```
RubenSousaDinis, line 20: this will not work. `talents` is already paginated. You need to fetch the info from the paging object

app/controllers/api/v1/public_api/talents_controller.rb
```diff
@@ -34,10 +34,11 @@ def index
   end
 
   def recommended
-    per_page = 3
-
     subscribed_users = Subscription.where(user_id: current_user.id)
```
RubenSousaDinis, line 37: deleted line being used below.

Also

```suggestion
    subscribed_users = Subscription.where(user: current_user)
```

PR #1201, opened 2023-05-17
Add quests endpoint

RubenSousaDinis:
## Summary

This PR adds an endpoint to fetch quests.

## Notion link

https://www.notion.so/talentprotocol/Earn-Page-Implementation-7e9e213bc9a04f3a90d11fc2f4284980?pvs=4#93b15cc5d7174bb9a387453c655d6bc1


app/controllers/api/v1/public_api/activities_controller.rb
```diff
@@ -0,0 +1,26 @@
+class API::V1::PublicAPI::ActivitiesController < API::V1::PublicAPI::APIController
```
RubenSousaDinis, line 1: @helloiambguedes when the controller is used to serialize more than one record it's a good practice to call name it in the plural. Almost all controllers are named in the plural.

app/controllers/api/v1/public_api/activities_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    pagy, activities = pagy_uuid_cursor(
+      total_activities,
+      before: cursor,
+      items: per_page,
+      order: {created_at: :desc, uuid: :desc}
+    )
+
+    response_body = {
+      activities: API::ActivityBlueprint.render_as_json(activities.includes([:origin_user, :target_user]), view: :normal),
```
RubenSousaDinis, line 16: When we're serializing associations we should include them. This way we're preventing N+1 queries.

app/controllers/api/v1/public_api/activities_controller.rb
```diff
@@ -0,0 +1,26 @@
+class API::V1::PublicAPI::ActivitiesController < API::V1::PublicAPI::APIController
+  before_action :internal_only
+  before_action :authenticated_only
+
+  def index
+    total_activities = current_user.activity_feed.all_activities
+
+    pagy, activities = pagy_uuid_cursor(
```
RubenSousaDinis, line 8: Our internal API uses cursor pagination. That's one of the reasons we need an uuid on API serialized records

spec/requests/api/v1/public_api/activities_spec.rb
```diff
@@ -1,45 +1,57 @@
...
-      let(:page) { 1 }
+      let(:cursor) { nil }
 
-      let!(:user) { create(:user, display_name: "API user") }
+      let!(:user) { create(:user, :with_talent_token, wallet_id: wallet_id, display_name: "API user") }
+      let(:wallet_id) { SecureRandom.hex }
+      let(:id) { wallet_id }
 
-      response "200", "Activities retrieved", save_example: true do
+      response "200", "Activities retrieved", document: false do
```
RubenSousaDinis, line 28: We use document false when we don't want to generate documentation for this test. This is useful for internal only API endpoints.

spec/requests/api/v1/public_api/activities_spec.rb
```diff
@@ -1,45 +1,57 @@
...
         before do
+          allow_any_instance_of(API::V1::PublicAPI::APIController).to receive(:current_user).and_return(user)
+
           career_updates = ["first", "second", "third", "fourth", "fifth", "sixth", "seventh", "eighth", "ninth"]
+          activity_feed = user.activity_feed
 
-          career_updates.each do |cu|
-            ac = Activities::CareerUpdate.create!(message: cu, origin_user_id: user.id)
+          career_updates.each do |message|
+            activity = create :activity, content: {message: message}.to_json, origin_user: user, type: "Activities::CareerUpdate"
```
RubenSousaDinis, line 51: It's a better practice to use factories in tests instead of business logic services

spec/requests/api/v1/public_api/activities_spec.rb
```diff
@@ -1,45 +1,57 @@
 require "swagger_helper"
 require "rails_helper"
 
-RSpec.describe "API::V1::PublicAPI::Activity", type: :request do
-  path "/activity" do
+RSpec.describe "Activities API", type: :request do
+  before do
```
RubenSousaDinis, line 5: this how we mock internal only API requests

app/blueprints/api/activity_blueprint.rb
```diff
@@ -1,6 +1,10 @@
 class API::ActivityBlueprint < Blueprinter::Base
   view :normal do
-    fields :created_at, :id, :content, :type
+    fields :created_at, :content, :type
+
+    field :id do |activity, _options|
```
RubenSousaDinis, line 5: API blueprints never serialize the internal database `id` only `uuid`

PR #1218, opened 2023-05-19
Add endpoint to fetch invites

RubenSousaDinis:
## Summary

This PR adds an endpoint to fetch user invites

## Notion link

https://www.notion.so/talentprotocol/Earn-Page-Implementation-7e9e213bc9a04f3a90d11fc2f4284980?pvs=4#d74dab07fa4449afa8b827cdd80fd7cc


app/blueprints/api/talent_blueprint.rb
```diff
@@ -78,4 +74,37 @@ class API::TalentBlueprint < Blueprinter::Base
...
+
+    field :joined_at do |user, _options|
+      user.created_at.to_s
+    end
+
+    field :participation_points_amount do |user, _options|
+      (user.talent&.verified? && user.created_at >= ParticipationPoints::CreditInvitePoints::START_DATE) ? ParticipationPoints::CreditInvitePoints::AMOUNT : 0
+    end
+
+    field :tal_amount do |user, _options|
```
RubenSousaDinis, line 89: This way we're backwards compatible with old rewards. We should not hide this info IMO

RubenSousaDinis: At least until we credit them in the Virtual Wallet

PR #1225, opened 2023-05-23
Disable stake and claim rewards when total supply is reached

verdefred:


app/packs/src/components/portfolio/NewPortfolio.jsx
```diff
@@ -505,7 +505,6 @@ const NewPortfolio = ({
           mode={theme.mode()}
           tokenAddress={tokenAddress}
           ticker={ticker}
-          onClaim={onClaim}
```
helloiambguedes, line 508: Don't we need this anymore?

helloiambguedes (approved):

app/packs/src/components/portfolio/NewPortfolio.jsx
verdefred: It was not being used, I just removed it

PR #1226, opened 2023-05-23
Add sponsorship volume metrics

francisco-leal:
## Summary

Adds a few extra metrics to track sponsorships: https://talentprotocol.notion.site/New-Data-Points-to-Metrics-Sheet-Sponsorship-Volume-and-Number-of-Users-a2d272f78fdc46b5b777d78e3690ba12


RubenSousaDinis (changes requested):
I think we're missing the upload to the excel sheet change?

PR #1262, opened 2023-06-05
Send notification after transaction commits

RubenSousaDinis:
## Summary

This PR fixes the invites notification since it's being sent inside a transaction which leads to an error in Sidekiq mailer.


app/services/users/create.rb
```diff
@@ -7,15 +7,15 @@ def initialize
     end
 
     def call(params)
-      ActiveRecord::Base.transaction do
-        invite = Invite.find_by(code: params[:code])
+      invite = invite_from(params[:code])
+      result = ActiveRecord::Base.transaction do
+        invite = invite_from(params[:code])
```
verdefred, line 12: can you explain why is this invite line repeated?

verdefred (approved):

PR #1289, opened 2023-06-19
fix: fix blank profile page

helloiambguedes:


app/packs/src/components/profile/overview/index.jsx
```diff
@@ -283,6 +283,8 @@ const Overview = ({
     }
   };
 
+  console.log(profile)
+
```
verdefred, line 287: extra console log

PR #1304, opened 2023-06-28
Quests V2

verdefred:
## Summary

Quests V2 include:
- 4 new quests
- Sorted quests per completeness
- "New" badge

## Notion link

https://www.notion.so/talentprotocol/Quests-Page-Improvements-714e82077a8b4e65ae9e074cd5b4dd05?pvs=4

## Notes

image


app/services/quests/refresh_user_quest.rb
```diff
@@ -119,6 +130,29 @@ def verify_humanity_quest_completed?
...
+      TalentSupporter.where(talent_contract_id: user.talent.talent_token.contract_id).count >= 3
+    end
+
+    def sponsor_talent_quest_completed?
+      return false unless user.wallet_id
+
+      Sponsorship.invested_by_user(user) >= 5
+    end
+
+    def invite_three_quest_completed?
```
RubenSousaDinis, line 152: Should this be after the release date? I think it would be better to incentivize more invites. Also, we should check the invite quality. Only count verified invites?

RubenSousaDinis: Also we are already crediting XP per invite so not sure why we want this quest

lib/tasks/quests.rake
```diff
@@ -61,6 +61,30 @@ namespace :quests do
...
+        experience_points_amount: 750
+      },
+      {
+        title: "Get 3+ Token Holders",
+        description: "Launch your Talent Token and get 3 people to stake on you.",
+        quest_type: "three_token_holders",
+        experience_points_amount: 2000
+      },
+      {
+        title: "Sponsor a Talent",
```
RubenSousaDinis, line 78: Isn't this quest an overlap with the `"Stake on 3+ Talent`? Everyone that completed that quest with complete this one... We're being redundant here in my opinion

config/sidekiq.yml
```diff
@@ -51,3 +51,9 @@
     update_chewy_index:
       cron: "0 * * * *" # Runs every hour
       class: TriggerChewyIndexUpdateJob
+    refresh_quests_job:
+      cron: "0 13 * * *" # Runs once every day at 13:00
+      class: Quests::RefreshQuestsJob
+    refresh_talent_mate_quests_job:
+      cron: "0 * * * *" # Runs every hour
+      class: Quests::RefreshTalentMateQuestsJob
```
RubenSousaDinis, line 59: In my opinion, this approach is not feasible. We're using an external API to fetch the user NFTs and call it for each verified user every hour. We will easily exceed the API tier we're currently using. IMO the best approach is to fetch the new-minted NFTs of the collection. Or have a call to action in the quest for the user to trigger the refresh.

RubenSousaDinis (commented):
Left some comments to discuss

config/sidekiq.yml
francisco-leal: Agreed. We have control over the backend that mints Talent Tokens, we can setup an endpoint to call it

The other option is to instead use The Graph, similar to the refresh supporters

verdefred: @RubenSousaDinis thanks for the comment, I was gonna ask you about API limits, since I didn't know. Having a CTA in the quest to trigger the refresh is a cool idea, I'm just not sure if it's feasible in terms of UI. I'll ask Filipe and PP if they have an idea for it and, if not, I'll go for fetching the new-minted NFTs of the collection and hoping they return the addresses of the owners

lib/tasks/quests.rake
verdefred: I don't think so. Staking and sponsoring are two different things. I can sponsor you with $5 but only stake in 1 person, or stake in 3 different people but not sponsoring anyone

app/services/quests/refresh_user_quest.rb
verdefred: Cool suggestion, I agree with the date and verification limits!
Regarding the crediting XP per invite, I think it's just an extra XP if you do it three times

config/sidekiq.yml
verdefred: Sorry @francisco-leal, not sure why but your comment was not showing up when I answered Rúben. I think that since we control the backend that mints Talent Tokens, maybe we can setup an endpoint to call it. That's probably the cleanest way, I'll take care of it, thanks!

helloiambguedes (approved):

PR #1346, opened 2023-07-31
feat: added approve button

helloiambguedes:


app/blueprints/api/talent_blueprint.rb
```diff
@@ -109,6 +109,10 @@ class API::TalentBlueprint < Blueprinter::Base
   view :overview do
     include_view :normal
 
+    field :wallet_address do |user, _options|
+      user.wallet_id
+    end
```
francisco-leal, line 114: We shouldn't need this because it's already included in the normal view

app/packs/src/components-v2/pages/profile/profile-header/index.jsx
```diff
@@ -203,6 +220,21 @@ export const ProfileHeader = ({ urlData, currentUser, isMobile }) => {
         profile={data.profileOverview}
         username={data.profileOverview.username}
       />
+      <ApprovalConfirmationModal
+        show={showApproveModal}
+        setShow={setShowApproveModal}
+        hide={() => setShowApproveModal(false)}
+        talent={data.profileOverview}
+        setProfile={noop}
```
francisco-leal, line 228: is noop really what we want to do? why not reload like below?

app/blueprints/api/talent_blueprint.rb
helloiambguedes: @verdefred let's talk a bit about this tomorrow

app/packs/src/components-v2/pages/profile/profile-header/index.jsx
helloiambguedes: Forcing the reload in the meantime until the edit flow is fixed

app/blueprints/api/talent_blueprint.rb
verdefred: Leal is correct, it shouldn't be needed

francisco-leal (approved):

PR #1361, opened 2023-08-04
Career Update - Edit and Delete functionalities

faustoxvx:
## Summary

Added the Edit and Delete functionality to career updates in the user's own Support page. Previously users could only add a new career update. Now users can Edit a previously made one and select/deselect goals to associate with that specific career update.

With the user experience in mind, the Delete functionality now has a delete-confirmation-modal. The typical "Are you sure..."(that can be reused for other parts of the application).

## Notion link

There was no notion link for this task.

## How to test?

Only functional tests at this point, but I added a video to demonstrate the flow working.

## Notes

Video walkthrough:

https://drive.google.com/file/d/1DsnprBuye7OlY46brOSj4wr6KlvqJOPm/view?usp=sharing


app/controllers/api/v1/public_api/career_updates_controller.rb
```diff
@@ -52,8 +52,38 @@ def create
...
+    Rollbar.error(
+      e,
+      "Unable to edit career update",
+      career_update_id: career_update&.id
+    )
+    render json: {error: "Unable to edit career update"}, status: :unprocessable_entity
+  end
+
+  def destroy
+    @career_update = CareerUpdate.find_by(uuid: params[:id])
```
francisco-leal, line 75: Replace this with `.find_by!`

app/controllers/api/v1/public_api/career_updates_controller.rb
```diff
@@ -52,8 +52,38 @@ def create
...
+    if @career_update.present?
+      @career_update.destroy
```
francisco-leal, line 77: Add a check if the destroy was successful or not and send the appropriate json message to the frontend

```rb
if @career_update.destroy
  render json: { ... }, status: :ok # there might be a better error code
else
  render json: { ... }, status: :unprocessable_entity
end
```

PR #1365, opened 2023-08-07
Pass chain id to backend for wallet

francisco-leal:


app/packs/src/components/virtual_portfolio/VirtualPortfolio.jsx
```diff
@@ -486,6 +489,7 @@ const VirtualPortfolio = ({ talent, railsContext }) => {
     const params = new URLSearchParams(document.location.search);
     const url = new URL(window.location.href);
     params.set("page", nextPage);
+    params.set("chain_id", chain_id);
```
verdefred, line 492: I think this should be params.set("chain_id", chainId);

verdefred (approved):

PR #1377, opened 2023-08-09
fix: recover verify flow

helloiambguedes:


app/packs/src/components-v2/pages/profile/profile-header/index.jsx
```diff
@@ -31,30 +31,38 @@ import { useImpersonate } from "./hooks/use-impersonate";
...
   const [showVerifyModal, setShowVerifyModal] = useState(false);
+  const [showWithPersonaModal, setShowWithPersonaModal] = useState(false);
   const dropdownMenu = useMemo(() => {
     if (!currentUser) return [];
     const menu = [{ value: "Share", iconColor: "primary01", iconName: "share-2" }];
     if (currentUser?.admin) {
       menu.push({ value: "Impersonate", iconColor: "primary01", iconName: "user" });
       if (!profileOverview?.verified) {
-        menu.push({ value: "Verify", iconColor: "primary01", iconName: "check" });
+        menu.push({ value: "Verify user", iconColor: "primary01", iconName: "check" });
```
verdefred, line 53: You need to add "profileOverview" to the dependencies array

app/packs/src/components-v2/pages/profile/profile-header/index.jsx
```diff
@@ -31,30 +31,38 @@ import { useImpersonate } from "./hooks/use-impersonate";
...
       }
       if (profileOverview?.profile_type === "waiting_for_approval") {
         menu.push({ value: "Approve", iconColor: "primary01", iconName: "check-chat" });
       }
     }
     if (currentUser?.username === urlData.profileUsername) {
       menu.push({ value: "Edit", iconColor: "primary01", iconName: "edit" });
+      if (!profileOverview?.verified && (
+        profileOverview?.with_persona_id ||
+        withPersonaRequest.requests_counter > railsContext.withPersonaVerificationsLimit)) {
```
verdefred, line 63: This is incorrect. You want to hide the option if the requests counter exceeds the limit or if the profile already has a with_persona_id (meaning they already uploaded a verification)

verdefred (commented):
Something is wrong in the process since if I click on the "Verify user" option and then click "continue" when selecting the legal name, my user gets automatically verified without needing to do the Persona flow

verdefred (approved):

PR #1378, opened 2023-08-09
Add virtual tal mint

francisco-leal:


db/migrate/20230809112707_add_migrated_to_rewards.rb
```diff
@@ -0,0 +1,5 @@
+class AddMigratedToRewards < ActiveRecord::Migration[7.0]
+  def change
+    add_column :rewards, :migrated, :boolean, default: false
```
verdefred, line 3: I would change this to a date migrated_at so we have more information

app/services/web3/mint_virtual_tal.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    def call(amount:, to:)
+      user = User.find_by(wallet_id: to)
+
+      return if user.nil?
+
+      decimals = "1#{"0" * 18}"
+      amount_to_charge = amount * decimals
+
+      tx_hash = client.transact_and_wait(tal_contract, "adminMint", to, amount_to_charge, 4, sender_key: key)
```
verdefred, line 20: Instead of 4, I would create the é um from our side so we are more explicit. What do you think?

PR #1393, opened 2023-08-16
feat: about tab ui upgrade

helloiambguedes:


app/packs/src/components-v2/pages/profile/profile-areas/about/index.jsx
```diff
@@ -15,14 +28,47 @@ export const About = ({ currentUser, urlData }) => {
...
+            leftIcon="edit"
+            onClick={openEditModal}
+          />
+        </ButtonContainer>
+      )}
+      {aboutData.career_goal.pitch && <AboutMe pitch={aboutData.career_goal.pitch} />}
+      {aboutData.current_position && (
+        <CurrentRole position={aboutData.current_position} username={urlData.profileUsername} />
+      )}
+      {aboutData.tags.length && <Tags tags={aboutData.tags} />}
```
francisco-leal, line 60: We have all these elements behind checks if they exist. But how does it look like when it's empty?

app/packs/src/components-v2/pages/profile/profile-areas/about/on-the-web/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+            </LinkItem>
+          );
+        }
+        return acc;
+      }, []),
+    [links]
+  );
+  return (
+    <Container>
+      <Typography specs={{ type: "medium", variant: "p1" }} color="primary01">
```
francisco-leal, line 42: This is going to look weird when there are no links

app/packs/src/components-v2/pages/settings/experiences/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+  useEffect(() => {
+    if (!username) return;
+    talentsService
+      .getMilestones(username)
+      .then(({ data }) => {
+        setMilestones(data.milestones);
+      })
+      .catch(() => {
+        console.error("Error fetching user milestones");
+        toast.error(<ToastBody heading="Error" body="There was a problem fetching user milestones." />);
```
francisco-leal, line 35: The user doesn't know what milestones are, can we rename this to experience?

app/packs/src/components-v2/pages/settings/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+        </DesktopColumn>
+        <DesktopColumn></DesktopColumn>
+      </DesktopPageContainer>
+    </>
+  );
+};
+
+export default (props, railsContext) => {
+  return () => (
+    <TalentThemeProvider>
```
francisco-leal, line 41: Do we still need this talent theme if we're not longer using light / dark modes?

config/routes.rb
```diff
@@ -134,6 +134,7 @@
   get "/u/:username/profile" => "profiles#preview"
   get "/u/:username" => "profiles#show", :as => "user", :constraints => {username: /[^\/]+/}
   get "/u/:username/account_settings" => "users#edit_profile", :as => "account_settings", :constraints => {username: /[^\/]+/}
+  get "/u/:username/more_info" => "profiles#settings", :as => "user_details", :constraints => {username: /[^\/]+/}
```
francisco-leal, line 137: I don't think this URL is the best, should we start using `settings`?

helloiambguedes: I'm not sure on this one

Checking another user experiences on the `/settings` tab is weird to me but I agree this route name is bad as well

app/packs/src/components-v2/pages/profile/profile-areas/about/index.jsx
helloiambguedes: It will only show the support data of that user
I can quickly add an empty state

verdefred (approved):

francisco-leal (approved):

PR #1410, opened 2023-08-25
feat: SendGrid contact lists

shirleyodeng:
## Summary

This PR;
- adds a rake task to import users as contacts to the SendGrid list `Members` (and `Talents Token V1` if they have a token)
- adds a job to run at 4am to delete all contacts on SendGrid
- adds a job to run at 5am to upsert all confirmed users as contacts on SendGrid

## Notion link

https://www.notion.so/talentprotocol/Bounty-Setup-Sendgrid-e260a94d1f8f408f9f5b549d27cafe70

## Notes

Will need to add these new environment variables in production:
- `ENABLE_SENDGRID_CONTACT_JOBS`
- `SENDGRID_MEMBERS_LIST_ID`
- `SENDGRID_TALENT_TOKENS_V1_LIST_ID`


lib/sendgrid/contacts/upsert.rb
```diff
@@ -0,0 +1,98 @@
...
+          fields[custom_field_id("Total_goals")] = total_goals if custom_field_id("Total_goals")
+          fields[custom_field_id("Total_updates")] = user.career_updates.size if custom_field_id("Total_updates")
+          fields[custom_field_id("Last_update")] = user.career_updates.order(created_at: :desc).first&.created_at if custom_field_id("Last_update")
+          # fields[custom_field_id("TAL_balance")] = user.created_at if custom_field_id("TAL_balance")
+          fields[custom_field_id("XP")] = user.experience_points_amount if custom_field_id("XP")
+          # fields[custom_field_id("Location")] = user.created_at if custom_field_id("Location")
+          fields[custom_field_id("Tags")] = user.tags.map(&:to_s).join(", ") if custom_field_id("Tags")
+          fields[custom_field_id("Subscribers")] = subscriptions.where(user:).size if custom_field_id("Subscribers")
+          fields[custom_field_id("Subscribing")] = subscriptions.where(subscriber: user).size if custom_field_id("Subscribing")
+          fields[custom_field_id("Friends_invited")] = User.where(invite_id: user.invites).size if custom_field_id("Friends_invited")
```
shirleyodeng, line 59: @francisco-leal

Can you check if these mappings are correct?

- Should `Joined` be `user.created_at` or `user.email_confirmed_at`?
- Where can I fetch `TAL_balance` and `Location` from?
- If the value is `nil`, should I not push the custom field to SendGrid or should I send a blank value eg. an empty string, except not sure if this would work for a date field?

francisco-leal: hey @shirleyodeng,

- Joined should be `user.email_confirmed_at`.
- location: `user.talent.location`, tal_balance: I've asked Filipe why he needs this field, I hope it will be enough to know if someone has TAL in their balance, because otherwise we'll need to keep running this Upsert service whenever there is a wallet interaction
- Sending a nil value should be ok 👍

shirleyodeng: @francisco-leal 

- Does this mean that I should only import users with a confirmed email to SendGrid?
- There is no `location` column on the talents table.
- Ah, Filipe has said not to send the field if it's blank.

app/models/user.rb
```diff
@@ -68,7 +68,9 @@ class User < ApplicationRecord
   # Elasticsearch index update
   update_index("talents", :talent)
 
+  after_destroy :delete_contact_on_sendgrid
   after_save :touch_talent
+  after_save :upsert_contact_on_sendgrid
```
shirleyodeng, line 73: @francisco-leal - is this okay to have these as callbacks or would you rather I enqueue the job all around the app?

lib/sendgrid/contacts/upsert.rb
francisco-leal: - Yes! If we send out emails to fake emails we'll get a bad score and might end up getting sent to spam folders
- We have a `profile` json field where we store some no so permanent fields (like location) there are attr_readers set up to make it easier to use
- Either way is fine, sendgrid should handle them the same way

francisco-leal: @shirleyodeng tagging you because I forgot on the previous comment

app/models/user.rb
francisco-leal: I think it's okay, we use `update_column` on the user table in some places where we do more recurring updates which doesn't trigger this (which is intended)

RubenSousaDinis (commented):
Let's get rid of the before and after hooks and have a single job that updates the contacts by deleting and creating them from scratch.

lib/sendgrid/contacts/upsert.rb
```diff
@@ -0,0 +1,124 @@
...
+        client = Eth::Client.create(ENV["CELO_RPC_URL"])
+        balance = client.call(tal_contract(ENV["VIRTUAL_TAL_ADDRESS_CELO"]), "getBalance")
+        balance.to_f.positive?
+      end
+
+      def polygon_balance?
+        client = Eth::Client.create(ENV["POLYGON_RPC_URL"])
+        balance = client.call(tal_contract(ENV["VIRTUAL_TAL_ADDRESS_POLYGON"]), "getBalance")
+        balance.to_f.positive?
+      end
```
shirleyodeng, line 109: @RubenSousaDinis - is this right? It doesn't look like it's unique to the user.

shirleyodeng: @RubenSousaDinis - thanks! Fixed!

.env
```diff
@@ -55,6 +55,8 @@ REDIS_URL=redis_url
 ROLLBAR_ACCESS_TOKEN=rollbar_access_token
 ROLLBAR_ENV=rollbar_env
 SENDGRID_API_KEY=sendgrid_api_key
+SENDGRID_MEMBERS_LIST_ID=sendgrid_members_list_id
+SENDGRID_TALENT_TOKENS_V1_LIST_ID=sendgrid_talent_tokens_v1_list_id
```
RubenSousaDinis, line 59: missing `ENABLE_SENDGRID_CONTACT_JOBS`

lib/sendgrid/contacts/delete.rb
```diff
@@ -0,0 +1,60 @@
+require "sendgrid-ruby"
+
+module Sendgrid
+  module Contacts
+    class Delete
+      class Error < StandardError; end
+
+      class BadResponse < Error; end
+
+      def initialize(emails: [])
```
RubenSousaDinis, line 10: Thanks for the flexibility 👍

.env
shirleyodeng, line 59:
```suggestion
SENDGRID_TALENT_TOKENS_V1_LIST_ID=sendgrid_talent_tokens_v1_list_id
ENABLE_SENDGRID_CONTACT_JOBS=enable_sendgrid_contact_jobs
```

shirleyodeng:
Replaced with https://github.com/talentprotocol/web-dapp/pull/1440

PR #1414, opened 2023-08-29
feat: settings page

helloiambguedes:


RubenSousaDinis:
When the username is updated we need to propagate the change to other profile components. Currently it generates some errors when we edit the username and click on a new profile tab.

RubenSousaDinis (approved):

PR #1435, opened 2023-09-05
feat: SendGrid dynamic templates

shirleyodeng:
## Summary

This PR uses SendGrid dynamic templates to send emails [noted here](https://www.notion.so/talentprotocol/New-Transactional-Emails-text-based-75f65983aebf4278a3859bf4f9b313c6).

## Notion link

https://www.notion.so/talentprotocol/Bounty-Setup-Sendgrid-e260a94d1f8f408f9f5b549d27cafe70


app/mailers/application_mailer.rb
```diff
@@ -4,6 +4,31 @@ class ApplicationMailer < ActionMailer::Base
...
+  end
+
+  def sendgrid_request_body(dynamic_template_data:, template_id:, to:)
+    {
+      dynamic_template_data:,
+      from: {
+        email: "no-reply@talentprotocol.com",
+        name: "Talent Protocol"
+      },
+      mail_settings: {sandbox_mode: {enable: !Rails.env.production?}},
```
shirleyodeng, line 26: Can't find ways to test this via the API except this: https://docs.sendgrid.com/for-developers/sending-email/sandbox-mode

RubenSousaDinis: Looks like the correct way for me!

app/mailers/application_mailer.rb
```diff
@@ -4,6 +4,31 @@ class ApplicationMailer < ActionMailer::Base
...
+  def sendgrid_api
+    @sendgrid_api ||= SendGrid::API.new(api_key: ENV["SENDGRID_API_KEY"])
+  end
+
+  def sendgrid_request_body(dynamic_template_data:, template_id:, to:)
+    {
+      dynamic_template_data:,
+      from: {
+        email: "no-reply@talentprotocol.com",
+        name: "Filipe at Talent Protocol"
```
RubenSousaDinis, line 24: I've created this env var so we can distinguish emails from prod and staging. I've added in both environments. Can you add it o the `.env` file?

```suggestion
        name: ENV['EMAILS_FROM']
```

shirleyodeng: Okay sounds good!

app/mailers/application_mailer.rb
```diff
@@ -4,6 +4,31 @@ class ApplicationMailer < ActionMailer::Base
 
   private
 
+  def send_sendgrid_email(dynamic_template_data:, template_id:, to:)
+    sendgrid_api
```
shirleyodeng, line 8: @RubenSousaDinis if the response is not successful, do you just want me to raise an exception with the response body?

RubenSousaDinis:
I think we can remove all email templates from the app in this pull request, no? I only see some of them being deleted

app/mailers/application_mailer.rb
shirleyodeng: Okay thanks, just can't seem to test this locally

shirleyodeng:
> I think we can remove all email templates from the app in this pull request, no? I only see some of them being deleted

@RubenSousaDinis not all emails are being transferred to SendGrid

app/mailers/application_mailer.rb
RubenSousaDinis: It doesn't trigger the http request? In the docs they mention it

shirleyodeng: It does but it just checks the arguments and doesn't test the email eg. I can't see what it looks like

PR #1437, opened 2023-09-06
Drop elasticsearch and replace search

francisco-leal:
## Summary

Removes the need to use elasticsearch in favour of a native solution for postgres.


app/controllers/api/v1/talent_controller.rb
```diff
@@ -3,40 +3,35 @@ class API::V1::TalentController < ApplicationController
...
-  end
-
-  # public /
-  def public_index
     talents =
-      if token_id_params.present?
-        Talent.includes(:talent_token, :user).where(talent_tokens: {contract_id: token_id_params})
+      if filter_params[:keyword].present?
+        query = <<~SQL
+          SELECT "users".* FROM "users"
```
RubenSousaDinis, line 9: I think you want a list of talents and not users here.

app/controllers/api/v1/talent_controller.rb
```diff
@@ -3,40 +3,29 @@ class API::V1::TalentController < ApplicationController
...
-    }, status: :ok
-  end
-
-  # public /
-  def public_index
     talents =
-      if token_id_params.present?
-        Talent.includes(:talent_token, :user).where(talent_tokens: {contract_id: token_id_params})
+      if filter_params[:keyword].present?
+        filter_query = ActiveRecord::Base.sanitize_sql_array(["SIMILARITY(COALESCE(display_name, username), :keyword) >= 0.3", keyword: "fran"])
```
RubenSousaDinis, line 8: hardcoded keyword

RubenSousaDinis, line 8: I think it would be cool to have this value configurable to test results easier

RubenSousaDinis: yes! But already understood 0.3 is the recommended setting. All good to keep it as is

RubenSousaDinis (approved):

verdefred (approved):

PR #1447, opened 2023-09-08
Associate goals with users

RubenSousaDinis:
## Summary

Associate user directly with goals

## Notion link

https://www.notion.so/talentprotocol/Write-Goal-Flow-V1-0-Dev-MVP-ed08131c7caa4dd39309aef570b536ba?pvs=4#9aa5ea479ab94c37864f42a6fb82ee1b


db/migrate/20230908133614_add_user_to_goals.rb
```diff
@@ -0,0 +1,5 @@
+class AddUserToGoals < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :goals, :user, foreign_key: true
```
verdefred, line 3: should we add an index?

RubenSousaDinis: reference already adds an index by default

RubenSousaDinis: check the schema

verdefred (approved):

PR #1450, opened 2023-09-11
fix: SendGrid delete contact

shirleyodeng:
## Summary

This PR changes the way we delete contacts from SendGrid by;
- running a job once a day to delete non members as contacts from SendGrid
- delete contact from SendGrid when a user is destroyed

## Notion link

https://www.notion.so/talentprotocol/Setup-Sendgrid-e260a94d1f8f408f9f5b549d27cafe70


RubenSousaDinis (approved):
Thanks!

PR #1451, opened 2023-09-11
Remove mandatory career goal from relationships

RubenSousaDinis:
## Summary

This PR removes the mandatory need of a `career_goal_id` in goals and career_needs. 

This is just a preparatory PR. We won't delete any data yet. In the next PR I will remove any usage of the `career_goal_id` from the app.

## Notion link

https://www.notion.so/talentprotocol/Write-Goal-Flow-V1-0-Dev-MVP-ed08131c7caa4dd39309aef570b536ba?pvs=4#41d729b0325a4a0b812a0a4664d27e61

### DIFF
--- a/app/models/career_need.rb
+++ b/app/models/career_need.rb
@@ -1,5 +1,5 @@
 class CareerNeed < ApplicationRecord
-  belongs_to :career_goal
+  belongs_to :career_goal, optional: true
 
   HIRING_NEEDS = ["Finding and hiring talent"].freeze
   ROLE_NEEDS = [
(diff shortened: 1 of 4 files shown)


app/models/career_need.rb
```diff
@@ -1,5 +1,5 @@
 class CareerNeed < ApplicationRecord
-  belongs_to :career_goal
+  belongs_to :career_goal, optional: true
```
verdefred, line 2: Should we keep the CareerNeed model, connected directly with the User or can we simply delete it?

RubenSousaDinis: I don't see them being used in the app. Are we using them?

verdefred: I don't think so. We can simply delete them but it doesn't need to be in this PR

verdefred (approved):

PR #1461, opened 2023-09-12
Add new goal endpoints and deprecate old endpoints that still use career goals

RubenSousaDinis:
## Summary

This PR adds new endpoints to the public API to update, create, and destroy goals.

It also adds deprecated notes to the existing goals endpoints.

In a next PR we will remove all existing `career_goal` calls from the client side.


app/controllers/api/v1/career_goals/goals_controller.rb
```diff
@@ -30,7 +31,6 @@ def create
     Rollbar.error(
       e,
       "Unable to create goal",
-      goal_id: goal.id,
       career_goal_id: career_goal.id
```
francisco-leal, line 34: I think you are deleting the wrong line

francisco-leal (approved):

PR #1481, opened 2023-09-20
Mint virtual TAL using lambda function

RubenSousaDinis:
## Summary

This PR adds a lambda call to mint virtual TAL


francisco-leal (approved):

francisco-leal (changes requested):
We need different lambdas for each environment

francisco-leal (approved):

PR #1488, opened 2023-09-25
fix: use goals association to user

shirleyodeng:
## Summary

This PR uses the new goals association to the user instead of going through talent and career goals.


RubenSousaDinis:
Ended up opening a new PR with all the fixes required. Other places had the same problem! Thanks for the help!

PR #1491, opened 2023-09-25
feat: career update streaks

shirleyodeng:
## Notion link

https://www.notion.so/talentprotocol/Career-Updates-Streaks-26da75d7fe914bdfa5f2f650cb8b5312


app/models/quest.rb
```diff
@@ -5,26 +5,32 @@ class Quest < ApplicationRecord
...
   validates :title, :quest_type, uniqueness: true
-  validates :experience_points_amount, :title, :description, :quest_type, presence: true
+  validates :title, :description, :quest_type, presence: true
   validates_inclusion_of :quest_type, in: VALID_QUEST_TYPES
 
+  has_many :quest_experience_points
   has_many :user_quests
 
+  def experience_points_amount
+    quest_experience_points.first&.amount || read_attribute(:experience_points_amount)
```
RubenSousaDinis, line 31: Do you have plans to remove `experience_points_amount` from the `quests`  table after merging this PR?

app/services/quests/calculate_monthly_update_streak_count.rb
```diff
@@ -0,0 +1,36 @@
+module Quests
+  class CalculateMonthlyUpdateStreakCount
```
RubenSousaDinis, line 2: I think we need to allow a start date to be passed to this service. If we have a bug we should be able to easily run a script that will update the progress accordingly for all users. We should only allow from October 2023 onwards tho

shirleyodeng: Okay, I'll need to hardcode this in the front-end too.

app/controllers/api/v1/public_api/quests_controller.rb
```diff
@@ -19,7 +22,21 @@ def index
...
     )
 
-    quests = quests.select("quests.*, user_quests.completed_at") if user
+    if user
+      quests =
+        quests
+          .select(
+            "
+              quests.*,
+              CASE
```
RubenSousaDinis, line 31: I'm not 100% convinced with this approach. I think we're making it hard to add new streak quests in the future with this approach. Would it be possible to create the `user_quest` even if the quest in not completed but in progress?

shirleyodeng: @RubenSousaDinis - sorry I realised I forgot to add some commits but I removed the unique index on user quests for user_id and quest_id. We will only validate this for non-streak quests as streak quests can have more than one as they will be created for each quarterly and yearly bonus so what you said is not possible as here we check if it's completed for the month, not quarter / year?

RubenSousaDinis (approved):

PR #1494, opened 2023-09-25
Add current user number of votes

verdefred:


francisco-leal (approved):

app/blueprints/api/user_blueprint.rb
```diff
@@ -27,6 +27,12 @@ class API::UserBlueprint < Blueprinter::Base
         Vote.where(candidate_id: user.id).sum(&:amount)
       end
     end
+
+    field :current_user_vote_count do |user, options|
+      if options[:with_vote_count]
```
RubenSousaDinis, line 32: where is this being used?

verdefred: app/blueprints/api/organization_blueprint.rb:64

PR #1520, opened 2023-09-28
Voting error handling

verdefred:
## Notion link

https://www.notion.so/talentprotocol/Rewards-Tab-Testing-Feedback-d52c860cd8594a7fa15edcc678aae439?pvs=4


app/controllers/api/v1/public_api/experience_rewards_controller.rb
```diff
@@ -13,10 +13,20 @@ def claim
...
-    end
+    render json: {claim: result}, status: :ok
+  rescue Rewards::Claim::NotVerified => e
+    render json: {error: e.message}, status: :conflict
+  rescue Rewards::Claim::OutOfStockReward => e
+    render json: {error: e.message}, status: :conflict
+  rescue Rewards::Claim::NotEnoughExperiencePoints => e
+    render json: {error: e.message}, status: :conflict
+  rescue Rewards::Claim::ErrorWhileGivingReward => e
+    render json: {error: e.message}, status: :conflict
```
RubenSousaDinis, line 24:
```suggestion
  rescue Rewards::Claim::Error=> e
    render json: {error: e.message}, status: :conflict
```

app/controllers/api/v1/public_api/experience_rewards_controller.rb
```diff
@@ -13,10 +13,20 @@ def claim
...
+  rescue ActiveRecord::RecordNotFound => e
```
RubenSousaDinis, line 25: 404 is already handle the by the API controller we can remove from here

verdefred: Since we have a final rescue, it isn't. it gives out 500 instead of 404

app/controllers/api/v1/public_api/experience_rewards_controller.rb
verdefred: Nice! I'll make the change

app/controllers/api/v1/public_api/experience_rewards_controller.rb
RubenSousaDinis: ok, that's why I don't like to rescue StandardError and send a non-500 response. All unexpected errors from the API should be a 500 since something is actually wrong with the code

RubenSousaDinis (approved):

PR #1521, opened 2023-09-28
ETHLisbon voting

verdefred:
## Notion link

https://www.notion.so/talentprotocol/ETHLisbon-Scholarships-323cbd29426e41dabf50286e28daaf77?pvs=4


db/migrate/20230928220752_add_links_to_elections.rb
```diff
@@ -0,0 +1,6 @@
+class AddLinksToElections < ActiveRecord::Migration[7.0]
+  def change
+    add_column :elections, :terms_and_conditions, :string
+    add_column :elections, :learn_more, :string
```
RubenSousaDinis, line 4: In case this is a URL:
```suggestion
    add_column :elections, :terms_and_conditions_url, :string
    add_column :elections, :learn_more_url, :string
```

app/blueprints/api/goal_blueprint.rb
```diff
@@ -12,6 +12,10 @@ class API::GoalBlueprint < Blueprinter::Base
       goal.election&.status
     end
 
+    field :election_slug do |goal, options|
```
RubenSousaDinis, line 15:
```suggestion
    field :organization_slug do |goal, options|
```

app/services/goals/create.rb
```diff
@@ -22,7 +22,7 @@ def call
       ActiveRecord::Base.transaction do
         raise CreationError unless goal.save
 
-        add_to_collective(goal) if params[:election_selected]
+        add_to_collective(goal, params[:election_selected]) if params[:election_selected]
```
RubenSousaDinis, line 25: I'm confused. I believe we should pass the election `uuid` and not the organization of the election

verdefred: Election has no uuid. For simplicity sakes I took this approach, to not change much of the code

app/blueprints/api/goal_blueprint.rb
verdefred: In this case it's the same thing. I want to associate this with the election, since it's what we have on goals, and not an organization

RubenSousaDinis (approved):

PR #1530, opened 2023-10-02
feat: add rake task to send reactivation email to legacy accounts

shirleyodeng:
## Summary

This PR adds a rake task to send reactivation email to legacy accounts.

## Notion link

https://www.notion.so/talentprotocol/Reactivate-Legacy-Accounts-Follow-up-Mail-238e3ed6f33a4144ad3efbb7f0d5baf3?d=3df1e7059f754d9d90f890155b4173f6


lib/tasks/users.rake
```diff
@@ -6,12 +6,21 @@ namespace :users do
...
     Talent.joins(:career_goal).where.not(career_goal: {pitch: nil}).find_each do |talent|
       talent.about = talent.career_goal.pitch
       talent.save(touch: false)
     end
   end
+
+  desc "Send reactivation email to legacy accounts"
+  task send_reactivation_email_to_legacy_accounts: :environment do
+    User
+      .where("last_sign_in_at IS NOT ? AND last_sign_in_at > ?", nil, 1.year.ago)
```
RubenSousaDinis, line 20: I don't think those fields are being updated anymore. We use `last_access_at` now: https://github.com/talentprotocol/web-dapp/blob/5825d1c7992d8feb0d1cb953c2ff9142a27099ae/app/controllers/application_controller.rb#L82-L86

shirleyodeng: @RubenSousaDinis ah I created a new user and it updated but I'll change it.

RubenSousaDinis (approved):

PR #1542, opened 2023-10-03
fix: add asm to SendGrid request

shirleyodeng:
## Summary

This PR adds `asm` to the SendGrid request to specify which unsubscribe groups to show.


RubenSousaDinis (approved):
Thanks!

PR #1557, opened 2023-10-06
Job when election ends

verdefred:
## Notion link

https://www.notion.so/talentprotocol/TOI-Deducting-TAL-onchain-and-distributing-prize-pool-bc0d5dcc697d48859202c8187f38aae0?pvs=4


app/services/elections/end_election.rb
```diff
@@ -0,0 +1,75 @@
...
+        .select("SELECT candidate_id, SUM(amount) as sum_votes")
+        .where(election_id: election.id)
+        .group(:candidate_id)
+        .order(Arel.sql("COALESCE(SUM(amount), 0) DESC"))
+        .limit(election.winners_count)
+        .pluck(:candidate_id)
+    end
+
+    def prize_pool
+      pluck_query = ActiveRecord::Base.sanitize_sql_array(["SUM(cost::float / 1000000000000000000)"])
```
RubenSousaDinis, line 59: instead of 1000000000000000000 let's use `TalentToken::TAL_DECIMALS` to increase readability

app/services/elections/end_election.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    private
+
+    attr_reader :election
+
+    def deduct_tal_for_votes
+      ::Vote.where(election_id: election.id).find_each do |vote|
+        service = Web3::BurnVirtualTal.new(chain_id: vote.chain_id)
+        tx = service.call(amount: vote.cost, from: vote.voter.wallet_id, reason: "vote")
```
RubenSousaDinis, line 28:
```suggestion
tx =  service.call(amount: vote.cost, from: vote.voter.wallet_id, reason: "vote", description: "Deducted from <organization name> voting")
```

app/services/elections/end_election.rb
```diff
@@ -0,0 +1,75 @@
...
+    def distribute_prize_pool
+      total_winning_votes = ::Vote.where(election_id: election.id, candidate_id: top_winners).sum(:amount)
+
+      voter_and_votes_count_pairs.each do |voter_id, votes_count|
+        amount_to_receive = prize_pool * 0.9 * votes_count / total_winning_votes
+        to = User.find_by(id: voter_id).wallet_id
+        next unless to
+
+        service = Web3::MintVirtualTal.new(chain_id: POLYGON_ID)
+        service.call(amount: amount_to_receive.to_s, to: to, reason: "winning_vote")
```
RubenSousaDinis, line 43:
```suggestion
        service.call(amount: amount_to_receive.to_s, to: to, reason: "winning_vote", description: "Earned from <organization name> voting prize pool")
```

app/jobs/deduct_tal_for_vote_job.rb
```diff
@@ -5,7 +5,7 @@ def perform(vote_id:, chain_id:)
     vote = Vote.find_by(id: vote_id)
 
     service = Web3::BurnVirtualTal.new(chain_id: chain_id)
-    tx = service.call(amount: vote.cost, from: vote.voter.wallet_id)
+    tx = service.call(amount: vote.cost, from: vote.voter.wallet_id, reason: "vote")
```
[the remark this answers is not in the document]
verdefred: It's not. This job was meant to be used if we burned TAL while voting. I'll delete it

app/services/elections/end_election.rb
```diff
@@ -0,0 +1,85 @@
...
+    private
+
+    attr_reader :election
+
+    def deduct_tal_for_votes
+      ::Vote.where(election_id: election.id).find_each do |vote|
+        service = Web3::BurnVirtualTal.new(chain_id: vote.chain_id)
+        tx = service.call(
+          amount: vote.cost,
+          from: vote.voter.wallet_id,
```
RubenSousaDinis, line 30: We need to use `vote.wallet_id` here. Some users might have swapped wallets in the meantime

app/services/elections/end_election.rb
```diff
@@ -0,0 +1,85 @@
...
+        tx ? vote.update!(tx_hash: tx) : vote.destroy
+      end
+    end
+
+    def distribute_prize_pool
+      total_winning_votes = ::Vote.where(election_id: election.id, candidate_id: top_winners).sum(:amount)
+
+      voter_and_votes_count_pairs.each do |voter_id, votes_count|
+        amount_to_receive = prize_pool * 0.9 * votes_count / total_winning_votes
+        to = User.find_by(id: voter_id).wallet_id
```
RubenSousaDinis, line 44: same as above. We should credit without double checking if the user exists

verdefred: I don't think we can do that here (or, at least, it's a bit weird) because we don't have a specific vote to get the wallet id, since the winner can have several votes.
maybe we can get the wallet_id from the first vote the user did instead of the wallet_id? But I don't know if that's better

RubenSousaDinis: Yes! you're right

app/services/elections/end_election.rb
```diff
@@ -0,0 +1,85 @@
...
+        tx = service.call(
+          amount: vote.cost,
+          from: vote.wallet_id,
+          reason: "vote",
+          description: "Deducted from voting on #{vote.candidate.name} for #{election.organization.name}"
+        )
+
+        tx ? vote.update!(tx_hash: tx) : vote.destroy
+      end
+    end
```
francisco-leal, line 37: I think this logic needs to be a lot more complex...

1. We should exclude votes that already have a TX, otherwise we can't run this job twice.
2. As far as I remember our lambda function is not waiting for the transaction to finish, so at this point we don't have a confirmation that it actually ran.
3. I think we need to use the ruby library to know if this TX was already mined and if so we check that the balance before & after actually make sense.

Now, talking about the `vote.destroy` we should destroy not after a TX is run, but if the user balance is lower than the amount we want to burn and even so in this case I'd use a "hidden" or "insufficient_balance" boolean or something like that so that we can manually check this. We're doing this first for EthLisbon so there is a lower amount of votes which means we should do this process almost manually.

When we delete a vote that means that everyone else that comes after will pay less as well, so we also need to be calculating how many votes someone has in real time (that were created before the vote that we are analysing)

app/services/elections/end_election.rb
francisco-leal: We need to award to the wallet that did the vote, not the user. If we know who won why don't we just use a simple query like `Vote.where(candidate_id: )` and then give the reward equal to the number of votes?

From the get go we already know how much each vote is worth, so we just need to multiply `vote.amount` by that.

We also need to know which Votes have already received their tokens worth or else we won't be able to run this job more than once and it will be messy to understand where and how it went wrong

verdefred: Having a transaction per winning vote seems redundant. Why do you feel that we need to award the wallet that did the vote and not the user?

francisco-leal: It's per per vote yes, but a vote can have a big amount so not like it's going to be a huge amount... To be honest, so that people aren't able to move TAL around between wallets. But also because it's what we mentioned here: https://discord.com/channels/834047763661455399/917833264616574986/1160832425623027722

verdefred: Got it, makes sense

app/services/wallet_activities/create.rb
```diff
@@ -0,0 +1,48 @@
...
+        chain_id: chain_id
+      )
+    end
+
+    private
+
+    attr_reader :wallet_activity
+
+    def upsert_on_chain(wallet_activity)
+      UpsertOnChain.new(wallet_activity: wallet_activity).call
```
verdefred, line 61: I think it's safer to do this async, since we don't know how much time the lambda function will take

app/jobs/wallet_activities/sync_latest_activities_job.rb
```diff
@@ -0,0 +1,9 @@
+class WalletActivities::SyncLatestActivitiesJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    WalletActivity.where(tx_succeeded_at: nil).where("onchain_syncs_count < 5").find_each do |wallet_activity|
```
verdefred, line 5: What happens when the onchain_syncs_count goes to 5? We have a WalletActivity that's not confirmed. Should that trigger some action?

db/migrate/20231009141925_add_new_fields_to_wallet_activities.rb
```diff
@@ -0,0 +1,15 @@
+class AddNewFieldsToWalletActivities < ActiveRecord::Migration[7.0]
+  def change
+    # Timestamp used to double check if a wallet activity was confirmed by our app
+    add_column :wallet_activities, :tx_succeeded_at, :timestamp
+    add_column :wallet_activities, :transfer_reason, :string
+    add_column :wallet_activities, :description, :string
+    add_column :wallet_activities, :readable_token_amount, :integer
```
verdefred, line 7: I think this should be a float. For voting we only allow for integers but that'll not be the case in other scenarios

app/jobs/wallet_activities/sync_latest_activities_job.rb
RubenSousaDinis: I will build a check on blazer that we can check when that happen

db/migrate/20231009174424_add_new_fields_to_votes.rb
```diff
@@ -0,0 +1,6 @@
+class AddNewFieldsToVotes < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :votes, :wallet_activity, foreign_key: true
```
verdefred, line 3: In this case we'll connect the burning of the vote with the wallet activity but I think it would be interesting to also connect the earning per vote. In that case it'll be easier to compare $TAL spent and $TAL won for all winners. wdyt?

app/jobs/wallet_activities/sync_latest_activities_job.rb
verdefred: Is there a way for us to get notified in case that happens? I think now, in the beginning, we'll be paying attention to it but after a while I'm afraid that we'll miss that

app/services/wallet_activities/create.rb
RubenSousaDinis: In most paths, this is already running inside of a job. I think only quests tal distribution is running sync. We would be creating lots of jobs

app/jobs/wallet_activities/sync_latest_activities_job.rb
RubenSousaDinis: Yes, we can setup notifications in Blazer

db/migrate/20231009141925_add_new_fields_to_wallet_activities.rb
RubenSousaDinis: I placed it as a decimal. I think it's better for currency values

RubenSousaDinis: ah got it, forgot to push

db/migrate/20231009174424_add_new_fields_to_votes.rb
RubenSousaDinis: We're doing that indirectly with the source relation in the wallet activity as well. We pass the object that led to the creation of the wallet activity. In case of token burns we pass the vote. In case of winning votes we pass the election

app/services/elections/end_election.rb
```diff
@@ -76,10 +80,10 @@ def prize_pool
 
     def voter_and_votes_count_pairs
       @voter_and_votes_count_pairs ||= ::Vote
-        .select("SELECT voter_id, SUM(amount) as sum_votes")
+        .select("SELECT wallet_id, SUM(amount) as sum_votes")
         .where(election_id: election.id, candidate_id: top_winners)
-        .group(:voter_id)
-        .pluck(:voter_id, "sum(amount)")
+        .group(:wallet_id)
```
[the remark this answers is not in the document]
RubenSousaDinis: Yes, you're right

app/services/wallet_activities/create.rb
verdefred: I think one is enough. We create a job to encapsulate the UpsertOnChain and that should work

app/services/elections/end_election.rb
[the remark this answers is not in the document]
RubenSousaDinis: Changed

app/services/wallet_activities/create.rb
RubenSousaDinis: It would be better to create the quest wallet activity inside of a job. I don't see the point of triggering many jobs.

- the distribution of election rewards already runs inside a job. 
- the distribution of tal from XP claim as well

francisco-leal (approved):

PR #1579, opened 2023-10-11
Add job to run a single election distribution

RubenSousaDinis:
## Summary

This PR adds a job to run a specific election distribution


app/jobs/elections/end_elections_job.rb
```diff
@@ -1,15 +0,0 @@
...
-      # winners_confirmation_period
-      elections = Election.where(voting_end_date <= Date.today && rewards_distributed == false)
-
-      elections.find_each do |election|
-        service = Elections::EndElection.new(election: election)
-        service.call
-      end
-    end
-  end
-end
```
verdefred, line 15: I think we should keep this, so we can have this working automatically

RubenSousaDinis: We can add it later. Now it's better to keep an eye and only trigger it manually. It also had a bug that's why I ended up changing it

verdefred (approved):

PR #1594, opened 2023-10-12
Add service to sync on chain data with wallet activities

RubenSousaDinis:
## Summary

This PR adds a service to sync on chain tal data with wallet activities


db/migrate/20231012133148_add_origin_to_wallet_activities.rb
```diff
@@ -0,0 +1,6 @@
+class AddOriginToWalletActivities < ActiveRecord::Migration[7.0]
+  def change
+    add_column :wallet_activities, :origin, :string, null: false, default: "in_app"
+    add_column :wallet_activities, :reason, :integer
```
verdefred, line 4: should we also add a "hidden" column? That way we can select specific activities that we want to hide, instead of having to hide all activities from a certain origin

app/services/wallet_activities/sync_on_chain.rb
```diff
@@ -0,0 +1,130 @@
...
+        event_type: event_name,
+        symbol: "TAL",
+        reason:,
+        user: User.find_by(wallet_id: wallet),
+        chain_id:
+      )
+    end
+
+    def get_logs(from_block_number, to_block_number)
+      provider.eth_get_logs(address: tal_address, fromBlock: from_block_number, toBlock: to_block_number)
```
verdefred, line 70: I think we should add a rescue for this, so we don't stop the sync but know where it failed and why

app/services/wallet_activities/sync_on_chain.rb
```diff
@@ -0,0 +1,130 @@
...
+
+      transaction_receipt(last_imported_wallet_activity.tx_hash)["result"]["blockNumber"].to_i(16)
+    end
+
+    def last_imported_wallet_activity
+      @last_imported_wallet_activity ||= WalletActivity.smart_contract.order(tx_date: :desc).first
+    end
+
+    def transaction_receipt(tx_hash)
+      provider.eth_get_transaction_receipt(tx_hash)
```
verdefred, line 94: I think we should add a rescue for this, so we don't stop the sync but know where it failed and why

app/services/wallet_activities/sync_on_chain.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    def block_timestamp(block_number)
+      block = provider.eth_get_block_by_number(block_number, false)
```
verdefred, line 98: I think we should add a rescue for this, so we don't stop the sync but know where it failed and why

app/services/wallet_activities/sync_on_chain.rb
RubenSousaDinis: As is it needs to fail. We basically get the latest timestamp imported to the app to start the process

db/migrate/20231012133148_add_origin_to_wallet_activities.rb
RubenSousaDinis: we can do it, yes

app/services/wallet_activities/sync_on_chain.rb
RubenSousaDinis: Check here: https://github.com/talentprotocol/web-dapp/pull/1594/files#diff-a77920c7c838ccb06bf8b9b6165a8aa84a6280a9cac8fd2cc07a6c2441387160R83

PR #1618, opened 2023-10-18
Change token price to 0.06

RubenSousaDinis:
## Summary

This PR changes token price to 0.06


francisco-leal (dismissed):
Looks good but we should probably sanity check this in dev

PR #1619, opened 2023-10-18
Add service to import staking rewards into the app

RubenSousaDinis:
## Summary

This PR imports the pending staking rewards into the app


app/services/web3/staking_rewards_sync.rb
```diff
@@ -0,0 +1,79 @@
...
+      @chain_id = chain_id
+    end
+
+    def call
+      supporter_wallets.each do |wallet|
+        wallet_pending_reward = WalletPendingReward.find_or_initialize_by(
+          wallet: wallet
+        )
+
+        pending_rewards_balance = pending_rewards_balance(wallet)
```
francisco-leal, line 20: this has the same name as the method, feels a bit confusing

app/services/web3/staking_rewards_sync.rb
```diff
@@ -0,0 +1,79 @@
...
+        if celo_chain?
```
francisco-leal, line 21: I assume we're running this for both chain right? because the same wallet can have interacted with both chains

RubenSousaDinis: Yes! We pass it as param

francisco-leal (approved):

PR #1627, opened 2023-10-19
Show organization image on election goals

verdefred:


app/services/goals/create.rb
```diff
@@ -57,9 +57,13 @@ def goal_create_message(goal)
...
-      params[:images].each do |image|
-        goal_image = goal.goal_images.build(image: image[:image_data].as_json)
-        goal_image.image_derivatives!
+      if params[:images] && params[:images].length > 0 && params[:election_selected].nil?
+        params[:images].each do |image|
+          goal_image = goal.goal_images.build(image: image[:image_data].as_json)
+          goal_image.image_derivatives!
+        end
+      else
+        goal.goal_images.build
```
RubenSousaDinis, line 66: I think the following will simplify the method:
```suggestion
      return goal.goal_images.create! if params[:election_selected].present?
      return unless params[:images] && params[:images].length > 0
      
      params[:images].each do |image|
          goal_image = goal.goal_images.build(image: image[:image_data].as_json)
          goal_image.image_derivatives!
      end
```

RubenSousaDinis: Also, I don't think build creates the record in the database, but I might be wrong

RubenSousaDinis: I see! All good

RubenSousaDinis (approved):

PR #1638, opened 2023-10-23
Sync balances daily

RubenSousaDinis:
## Summary

This PR adds a service to sync local TAL balance on chain


app/controllers/api/v1/wallet_activities_controller.rb
```diff
@@ -3,7 +3,6 @@ class API::V1::WalletActivitiesController < ApplicationController
 
   def index
     wallet_activities = current_acting_user.wallet_activities
-      .confirmed
```
francisco-leal, line 6: This makes it so that we show the TX even before it runs right?

app/services/wallet_activities/refresh_local_tal_balances.rb
```diff
@@ -1,13 +1,18 @@
 module WalletActivities
-  class RefreshTalBalances
+  class RefreshLocalTalBalances
+    def initialize(wallet: nil)
+      @wallet = wallet
+    end
+
     def call
-      WalletTalBalance.delete_all
+      WalletTalBalance.delete_all unless wallet
```
francisco-leal, line 8: Why do we want to delete all? so that we refresh all balances?

app/services/web3/mint_virtual_tal.rb
```diff
@@ -6,30 +6,21 @@ class Error < StandardError; end
...
-      talent_tokens_sold: 3,
-      in_app_rewards: 4,
-      investor: 5,
-      experience_rewards: 4,
-      winning_vote: 4
-    }
-
-    def initialize(wallet_activity:)
-      @wallet_activity = wallet_activity
+    MINT_REASON = 4
```
francisco-leal, line 9: Can we rollback this? and instead have a default value that we can override?

app/controllers/api/v1/wallet_activities_controller.rb
RubenSousaDinis: No, we're only setting the tx after the batch is completed. No tx will be displayed until the balance is broadcasted

app/services/web3/mint_virtual_tal.rb
RubenSousaDinis: Not sure why we would want that. Can you provide an example?

francisco-leal: For the talent tokens follow-up we'll want to use reason 3, we might still have a new round of investment where we want to use 5, etc..

RubenSousaDinis: But this is a batch of activities. We can have a mix of them in the same batch. Going forward the reason is simply to sync the balance of the app. We still store the `transfer_reason` in the wallet activity and we will use it to display the data in the client.

app/controllers/api/v1/wallet_activities_controller.rb
RubenSousaDinis: No, wallet activities are created by the app when an action is performed (claim talent tokens, claim pending rewards ,quests, etc.). Then, every day at 00h, we sync the app balance with the smart contract.

PR #1672, opened 2023-10-27
feat: request to join collectives

shirleyodeng:
## Summary

This PR adds the following to the collective pages;
- a button, "Request to Join" for users to join a collective. They can also cancel this request.
- a new tab, "Pending Requests" for admins to accept/reject a request. It will create a notification / send an email on acception/rejection.
- adds a button, "Add member" for admins to add users to a collective.
- adds a button, "Remove" for each member for admins to remove a member from a collective.

## Notion link

https://www.notion.so/talentprotocol/Collective-Pages-Request-Join-0ed53bd64b3f41369400f1399b1cb7d2


app/controllers/api/v1/public_api/memberships_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def create
+    created_membership = Memberships::Create.new(params: membership_params, user: current_user).call
+
+    response_body = {membership: API::MembershipBlueprint.render_as_json(created_membership, view: :normal)}
+    log_request(response_body, :created)
+    render json: response_body, status: :created
+  rescue => error
+    Rollbar
```
RubenSousaDinis, line 26: I think this is business logic and should be inside the service

app/controllers/api/v1/public_api/memberships_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  def update
+    updated_membership = Memberships::Update.new(membership:, params: update_params, user: current_user).call
+
+    response_body = {membership: API::MembershipBlueprint.render_as_json(updated_membership, view: :normal)}
+    render json: response_body, status: :ok
+  rescue Memberships::Update::UnauthorizedError
+    unauthorized_request("You don't have access to perform that action")
+  rescue => error
+    Rollbar
+      .error(
```
RubenSousaDinis, line 46: I think this is business logic and should be inside the service

app/controllers/api/v1/public_api/memberships_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def destroy
+    Memberships::Destroy.new(membership:, user: current_user).call
+
+    render json: {}, status: :ok
+  rescue Memberships::Destroy::UnauthorizedError
+    unauthorized_request("You don't have access to perform that action")
+  rescue => error
+    Rollbar
```
RubenSousaDinis, line 63: I think this is business logic and should be inside the service

app/controllers/api/v1/public_api/memberships_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+        "Error updating membership",
+        current_user_id: current_user.id,
+        membership_params: membership_params.to_h
+      )
+
+    render json: {error: "Unable to update membership."}, status: :unprocessable_entity
+  end
+
+  def destroy
+    Memberships::Destroy.new(membership:, user: current_user).call
```
RubenSousaDinis, line 57: Add optional note for the admin to provide the reason why it was not accepted

RubenSousaDinis: Hey @shirleyodeng. Yes! We will pay for the extra time! I will put a comment on the task. We don't need to store the note. We can pass it to the notification and store it as an extra param, so you can keep the logic as is, but propagate the note to the notification.

RubenSousaDinis (approved):

PR #1700, opened 2023-11-08
Review Invite 3+ Friends quest

verdefred:
## Summary

- Check if there are three invited friends everytime someone is verified (no matter how)
- Increase the XP reward to 2,000 XP
- Clean the quest for all core team members

## Notion link

https://www.notion.so/talentprotocol/Review-Invite-3-Friends-Quest-d59a10170a9444fba53535a84d49157d?pvs=4


lib/tasks/quests.rake
```diff
@@ -149,4 +149,37 @@ namespace :quests do
...
+  task update_friends_invited_quest: :environment do
+    quest = Quest.find_by(quest_type: "invite_three")
+    quest_experience_points = quest.quest_experience_points.first
+    completed_at = Time.current
+
+    puts "Increasing reward to 2,000 XP..."
+    quest_experience_points.update!(amount: 2000)
+    puts "Done!"
+
+    puts "Cleaning quest to all core team members..."
```
RubenSousaDinis, line 162: Not sure I understood what we mean by cleaning

RubenSousaDinis: We're creating new records not removing them

verdefred: Put them as done

RubenSousaDinis (approved):

RubenSousaDinis (changes requested):
Just remembered that we need to change the upsert quests task with the new XP amount

verdefred:
> Just remembered that we need to change the upsert quests task with the new XP amount

Good catch, fixed

RubenSousaDinis (approved):

PR #1703, opened 2023-11-08
feat: add leave goal dropdown option

shirleyodeng:
## Summary

This PR adds a "Leave Goal" dropdown option to the goals page for users that are accountability buddies.

Screenshot 2023-11-08 at 22 39 49

## Discord link

https://discord.com/channels/834047763661455399/1168580364546084894/1170302755097825311


app/controllers/api/v1/public_api/goals_controller.rb
```diff
@@ -94,4 +94,19 @@ def goal_params
...
     )
   end
+
+  def updatable_goal
+    @updatable_goal ||=
+      Goal
+        .left_joins(:goal_accountability_users)
+        .where("goals.user_id = :user_id OR goal_accountability_users.user_id = :user_id", user_id: current_user&.id)
+        .distinct
+        .find_by!(uuid: params[:id])
```
RubenSousaDinis, line 104: Wdyt about this?

```suggestion
@updatable_goal ||= GoalAccountabilityUser
  .joins(:goal)
  .where!(user: current_user)
  .find_by!(goals: {uuid: params[:id]})&.goal
```

app/packs/src/components-v2/goal-card/index.jsx
```diff
@@ -41,6 +38,24 @@ export const GoalCard = ({ goal, openAddGoalModal, isOwner, username, reloadGoal
               console.error(err);
             });
           return;
+        case "Leave goal":
+          if (!currentUser) return;
+
+          goalsService
+            .updateGoal(goal.id, {
+              goal: {
+                accountability_users: goal.accountability_users.map(user => user.id).filter(id => id !== currentUser.id)
```
RubenSousaDinis, line 47: I think here we can only accept the ID of the user being removed. Otherwise, someone part of a goal can remove all buddies if he uses the API directly. Actually, I think we can't accept ids and we should only use the current user in the backend.

RubenSousaDinis: We can have a fake param. `remove_current_user: true` And in the backend if it's set we compute the new `accountability_users` array without the current user. WDYT?

app/controllers/api/v1/public_api/goals_controller.rb
RubenSousaDinis: I see. Ok keep it as is

app/packs/src/components-v2/goal-card/index.jsx
RubenSousaDinis: Yup, that's the better approach

RubenSousaDinis (approved):

PR #1704, opened 2023-11-08
feat: current location tasks

shirleyodeng:
## Summary

This PR;
- adds a quest for users to add their current location and nationalities
- adds a rake task to populate their current country from the previous current location field

## Notion link

https://www.notion.so/talentprotocol/Add-location-nationality-to-profile-332cc60f205b49e88f88367e32ecc913


app/models/quest.rb
```diff
@@ -3,6 +3,7 @@ class Quest < ApplicationRecord
     active_goal
     complete_profile
     connect_wallet
+    country
```
RubenSousaDinis, line 6: I would call it `complete_location`. Since country only matches one field

RubenSousaDinis (approved):

PR #1806, opened 2024-01-01
clean-up: remove Christmas quests

shirleyodeng:
## Summary

This PR removes the Christmas quests.

Screenshot 2024-01-01 at 19 45 42


RubenSousaDinis:
> @francisco-leal @RubenSousaDinis [this spec](https://github.com/talentprotocol/web-dapp/blob/dev/spec/services/web3/whitelist_talent_mate_spec.rb) is failing because:
> 
> ```ruby
> module Web3
>   class WhitelistTalentMate
>     def call(user:, level:)
>       return false if Time.current.year >= 2024
> ```

Code removed

RubenSousaDinis (approved):

PR #1847, opened 2024-01-22
Feat/goals week four

shirleyodeng:
## Summary

Screenshot 2024-01-22 at 22 46 53

## Notion link

https://www.notion.so/talentprotocol/Goals-App-Tech-e7feb4875c0f46bc8af56071efcdbbb7?pvs=4


app/models/quest.rb
```diff
@@ -51,6 +51,15 @@ class Quest < ApplicationRecord
     three_token_holders
     verify_humanity
     verify_identity
```
shirleyodeng, line 53: @RubenSousaDinis might be good to see how we can clean up old quests.. unfortunately we can't remove because they're linked to user quests? Unless I can delete those too?

RubenSousaDinis: I think we will delete them yes. But we will do it after the launch. I will make sure we have a task to clean te remainders after te go live

PR #1872, opened 2024-01-31
Beta release 31.Jan.2024

francisco-leal:
* Small fixes (#1871)


app/blueprints/api/user_blueprint.rb
```diff
@@ -15,11 +15,11 @@ class API::UserBlueprint < Blueprinter::Base
     end
 
     field :verified do |user, _options|
-      user.talent&.verified?
+      user&.talent&.verified?
```
RubenSousaDinis, line 18: hmm, no user in the User blueprint?

RubenSousaDinis: on those pages we should not arrive here. Can you share an example?

francisco-leal: Goals & Updates, because the current user is being passed to the frontend through the blueprint. I added a work around on the controller directly because you are right, we shouldn't even call the blueprint

PR #1905, opened 2024-02-12
fix: profile_completeness calculations

shirleyodeng:
## Summary

This PR fixes the profile_completeness calculations.

Criteria:
- Display Name
- Short Bio
- At least 1 tag
- At least 1 socia link
- Nationality? not sure about this one
- At least 1 career update
- At least 1 goal
- PFP -> Being this mandatory


lib/tasks/users.rake
```diff
@@ -46,4 +46,11 @@ namespace :users do
         talent.update_column(:country, country)
       end
   end
+
+  desc "Update profile completeness"
+  task update_profile_completeness: :environment do
+    User.find_each do |user|
```
shirleyodeng, line 52: @RubenSousaDinis - should this be only `User.members`?

RubenSousaDinis: It's ok for all

RubenSousaDinis (approved):

PR #1923, opened 2024-02-19
fix: counts for subscribing and subscribers

shirleyodeng:
## Summary

This PR fixes the count for subscribers and subscribing.


shirleyodeng:
@RubenSousaDinis are connections still being used?

RubenSousaDinis (approved):

PR #1937, opened 2024-02-26
fix: twitter card for home page

shirleyodeng:
## Summary

This PR fixes the Twitter card for the homepage.

Screenshot 2024-02-26 at 10 12 03

## Notion link

https://www.notion.so/talentprotocol/Change-Metadata-for-Twitter-3ce620db61bf45fa9a02ee1d285f6f14


app/helpers/meta_tags_helper.rb
```diff
@@ -10,7 +10,8 @@ def meta_description
   end
 
   def meta_image
-    content_for?(:meta_image) ? content_for(:meta_image) : DEFAULT_META["meta_image"]
+    meta_image = content_for?(:meta_image) ? content_for(:meta_image) : DEFAULT_META["meta_image"]
+    meta_image.starts_with?("http") ? meta_image : image_url(meta_image)
```
shirleyodeng, line 14: Needs to be an absolute path.

app/views/profiles/show.html.erb
```diff
@@ -5,7 +5,6 @@
   <% end %>
   <% if @talent["profile_picture_url"].present? %>
     <% content_for :meta_image, @talent["profile_picture_url"] %>
-    <% content_for :meta_card_type, "summary" %>
```
shirleyodeng, line 8: This was a duplicate type so it was showing "summarysummary"

config/routes.rb
```diff
@@ -77,12 +77,12 @@
     get "/join(/:invite_code)" => "auth#sign_up", :as => :sign_up
     get "/" => "landing#index", :as => :root
     get "/sign_in" => "auth#new", :as => :sign_in
-    get "/home" => "homepage#index", :as => :user_root
     get "/auth/forgot_passord" => "auth#forgot_password", :as => "forgot_password"
     get "/confirm_email(/:token)" => "auth#confirm_email", :as => "confirm_email"
     get "/auth/password" => "auth#reset_password", :as => "reset_password"
   end
 
+  get "/home" => "homepage#index", :as => :user_root
```
shirleyodeng, line 85: @RubenSousaDinis moved this outside of the format constraints, that ok?

RubenSousaDinis: We can probably do it for all endpoints, no?

shirleyodeng: @RubenSousaDinis it looks like it was intentionally done like this in the goals app PR so I wasn't sure? I'm not sure the other endpoints are share worthy so up to you.

RubenSousaDinis: Let's do it yes. Because some of them can also be shared on socials

francisco-leal (approved):

